PR #2, opened 2023-09-07
adds asf yaml file

keith-turner:


phrocker (approved):
LGTM based on what I read at https://cwiki.apache.org/confluence/pages/viewpage.action?spaceKey=INFRA&title=git+-+.asf.yaml+features

.asf.yaml
```diff
@@ -0,0 +1,39 @@
...
+  description: "Apache Accumulo Access Control Library"
+  homepage: https://accumulo.apache.org
+  labels:
+    - accumulo
+    - big-data
+    - hacktoberfest
+  features:
+    wiki: false
+    issues: true
+    projects: false
```
ctubbsii, line 32: It's useful to use projects. In the main repo, we use projects to track milestones for multiple versions still being developed (since GitHub doesn't support tagging multiple milestones). We probably won't use that here... we can keep one linear set of releases and just use the normal milestone feature. However, we also use projects for other things, and that might end up being useful in this repo as well. I would go ahead and enable it since it's not hurting anything to have it enabled and there's a good chance we'll end up using it.

```suggestion
    projects: true
```

.asf.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+notifications:
+  commits:      commits@accumulo.apache.org
+  issues:       notifications@accumulo.apache.org
+  pullrequests: notifications@accumulo.apache.org
+  jobs:         notifications@accumulo.apache.org
+  discussions:  notifications@accumulo.apache.org
```
ctubbsii, line 39: The job notifications could get a little spammy, and I don't think they are particularly useful in the mailing list archive. I would suggest dropping that. Unlike some of the others, I don't think it defaults to dev@... it just doesn't get used at all.

The discussions feature isn't enabled, so that setting is pointless to set... it's just using up space. I'd drop that to minimize the file and reduce confusion of what effect is it having.

```suggestion
```

PR #4, opened 2023-09-08
Use ANTLR4 for specification validation

dlmarion:


dlmarion:
I think there is more to do here in the tests, but wanted to get something up sooner rather than later.

ctubbsii (changes requested):
The context of these changes seem okay, but the indentation in the POM is a bit wonky. The pom needs to be run through the sortpom plugin. This repo doesn't quite have the quality checks for GitHub Actions yet to catch that stuff. If you want to merge this now, and we can fix the formatting / style with some build automation and GitHub Actions quality checks later, that would be fine.

dlmarion:
@ctubbsii - I think @keith-turner might be looking for your input / changes on the pom overall.

ctubbsii:
> @ctubbsii - I think @keith-turner might be looking for your input / changes on the pom overall.

I will definitely go spelunking in the POM and GitHub Actions to polish up anything prior to release... but for now, I don't think it's a particularly high priority while the effort is in its initial rapid prototyping and development. If you think it's holding something up, I can definitely prioritize it, though.

dlmarion:
In c43e905 I modified the ANTLR grammar to modify the parse tree such that all of the tokens in an OR and AND expression were below that node in the tree. Added test that performs all of the evaluations in testdata.json. At this point the ANTLR expression parsing tests and AccessExpression evaluation tests are complete and working.

pom.xml
```diff
@@ -82,6 +82,8 @@
             <phase>generate-sources</phase>
             <configuration>
               <outputDirectory>target/generated-sources/antlr4/org/apache/accumulo/access/grammars</outputDirectory>
```
ctubbsii, line 84: What's wrong with the default output directory at: `${project.build.directory}/generated-sources/antlr4`?

src/main/java/org/apache/accumulo/access/AccessEvaluatorImpl.java
```diff
@@ -30,7 +30,7 @@
 import java.util.stream.Stream;
 
 //this class is intentionally package private and should never be made public
-class AccessEvaluatorImpl implements AccessEvaluator {
+public class AccessEvaluatorImpl implements AccessEvaluator {
```
ctubbsii, line 33: This change conflicts with the comment right above it.

ctubbsii (changes requested):
It feels very weird to write the specification in ABNF, and then to have a test to verify that the ABNF is properly written using an imported G4 specification for ABNF (which is imported to our package namespace, which doesn't need to be). And then, on top of that, have an entirely separate G4 specification that is basically redundant with our ABNF spec.

Why not just write the specification in G4 directly and drop all the ABNF stuff? We don't need to define the specification in two separate languages. We should just have one authoritative one.

pom.xml
dlmarion: IIRC I ended up with this because the output generated java classes with the correct package name, but put them in that directory. It didn't create the correct directory structure. I can re-test this to be sure. I think this issue points to the same problem: https://github.com/antlr/antlr4/issues/4202

src/main/java/org/apache/accumulo/access/AccessEvaluatorImpl.java
dlmarion: Yes, I can fix by making a method on the interface that calls this method as is done with some of the other methods. I wanted @keith-turner to look at this first though to be sure that I got the escaping/quoting correct in AccessExpressionEvaluator.

dlmarion:
> It feels very weird to write the specification in ABNF, and then to have a test to verify that the ABNF is properly written using an imported G4 specification for ABNF (which is imported to our package namespace, which doesn't need to be). 

I wanted to verify that our ABNF was correct and I used ANTLR to do it since it already has a grammar written for it. Our ABNF in the main branch is not correct and there are updates to it in this PR.

> And then, on top of that, have an entirely separate G4 specification that is basically redundant with our ABNF spec. 
> Why not just write the specification in G4 directly and drop all the ABNF stuff? We don't need to define the specification in two separate languages. We should just have one authoritative one.

I think we can delay the decision to use ABNF or ANTLR as the specification grammar until we determine what we are going to do with the ANTLR implementation (whether the ANTLR implementation is suitable to be our runtime implementation or just an example that others can leverage).

dlmarion:
Results from running benchmark from 2971c0d:
```
Benchmark                                                    Mode  Cnt  Score   Error   Units
AccessExpressionAntlrBenchmark.measureBytesParsing          thrpt   12  0.829 ± 0.012  ops/us
AccessExpressionAntlrBenchmark.measureEvaluation            thrpt   12  1.995 ± 0.013  ops/us
AccessExpressionAntlrBenchmark.measureEvaluationAndParsing  thrpt   12  0.519 ± 0.022  ops/us
AccessExpressionAntlrBenchmark.measureStringParsing         thrpt   12  0.851 ± 0.025  ops/us
```

Certainly not as good as the results listed in #7 with the old or new implementation.

pom.xml
ctubbsii: I noticed the directory structure issue when I ran it, but it didn't seem to matter for the build. Since the files aren't checked in from the target directory, it probably doesn't matter.

src/main/java/org/apache/accumulo/access/AccessEvaluator.java
```diff
@@ -56,6 +55,17 @@ public interface AccessEvaluator {
...
+   * Properly escapes an authorization string. The string can be quoted if desired.
+   *
+   * @param auth
+   * @param quote
+   * @return
+   */
+  static byte[] escape(byte[] auth, boolean quote) {
+    return AccessEvaluatorImpl.escape(auth, quote);
+  }
+
```
keith-turner, line 68: I think this can be removed.  I have another comment that changes the code using this to an existing API method.

```suggestion
```

src/test/java/org/apache/accumulo/access/grammar/antlr/AccessExpressionAntlrEvaluator.java
```diff
@@ -0,0 +1,119 @@
...
+    entities = new ArrayList<>(authSets.size());
+
+    for (Authorizations a : authSets) {
+      Set<String> entityAuths = a.asSet();
+      Entity e = new Entity();
+      entities.add(e);
+      e.authorizations = new HashSet<>(entityAuths.size() * 2);
+      a.asSet().stream().forEach(auth -> {
+        e.authorizations.add(auth);
+        e.authorizations.add(new String(AccessEvaluator.escape(auth.getBytes(UTF_8), true), UTF_8));
```
keith-turner, line 48: Can use this existing API method.

```suggestion
        var quoted = AccessExpression.quote(auth);
        if (!quoted.startsWith("\"")) {
          quoted = '"' + quoted + '"';
        }
        e.authorizations.add(quoted);
```

src/test/java/org/apache/accumulo/access/grammar/antlr/Antlr4Tests.java
```diff
@@ -0,0 +1,203 @@
...
+    // using ANTLR have the same outcome as AccessEvaluatorImpl
+    List<TestDataSet> testData = AccessEvaluatorTest.readTestData();
+    for (TestDataSet testSet : testData) {
+      for (TestExpressions test : testSet.tests) {
+        for (String expression : test.expressions) {
+          List<Authorizations> authSets =
+              Stream.of(testSet.auths).map(Authorizations::of).collect(Collectors.toList());
+          AccessEvaluator evaluator = AccessEvaluator.builder().authorizations(authSets).build();
+
+          AccessExpressionAntlrEvaluator antlr = new AccessExpressionAntlrEvaluator(authSets);
```
keith-turner, line 131: These can be moved outside of the loop.

```suggestion
        List<Authorizations> authSets =
              Stream.of(testSet.auths).map(Authorizations::of).collect(Collectors.toList());
        AccessEvaluator evaluator = AccessEvaluator.builder().authorizations(authSets).build();

        AccessExpressionAntlrEvaluator antlr = new AccessExpressionAntlrEvaluator(authSets);
        
        for (String expression : test.expressions) {
```

keith-turner (commented):
Since we have determined using ANTLR is not suitable for the impl of the java library, thinking we should move the antlr code out of java library code into its own directory.  Thinking it would be good to have a documentation directory and it could go there and be linked from the readme.

```
docs/SPECIFICATION.md     # move specification to new docs dir
docs/GETTING_STARTED.md   # this does not exists, but would give instructions for someone who wants to use the java library and simple examples of using the library... helps someone quickly use the java library.
docs/antlr/README.md      # a readme that goes over the different parts of the antlr example
docs/anrlt/pom.xml        # a pom for building the antlr example
docs/antlr/src            # source for the antlr example
```

Thinking overall we have a few different audiences for the above documentation.

 1.  People who want to understand this project at a conceptual level, hopefully SPECIFICATION.md fills that need.
 2. People who want to use the java library, hopefully GETTING_STARTED.md, the libraries javadoc, and SPECIFICATION.md get them on their way
 3. People who want to implement this in other languages.  For this SPECIFICATION.md, antlr docs, and the existing java source code would be useful.
 4. People who want to experiment with the language, for example compile an access expression to WASM or javascript.  For this activity, the antlr grammar would be really really useful in helping someone get started and quickly experimenting.

For cases 3 and 4 above, the main thing someone would need is the antlr grammar.  Any java code on top of that should serve the documentation use case of helping someone getting started doing whatever they want to do with the language.

src/main/java/org/apache/accumulo/access/AccessEvaluator.java
dlmarion: Implemented change in 6fd8db5

src/test/java/org/apache/accumulo/access/grammar/antlr/AccessExpressionAntlrEvaluator.java
dlmarion: Implemented change in 6fd8db5

src/test/java/org/apache/accumulo/access/grammar/antlr/Antlr4Tests.java
dlmarion: Implemented change in 6fd8db5

.gitignore
```diff
@@ -34,3 +35,16 @@
 /nb-configuration.xml
 /.vscode/
 /.factorypath
+/contrib/antlr4/.settings/
```
keith-turner, line 38: could create a .gitignore file in ./contrib/antlr4  and put these lines there

contrib/antlr4/src/test/java/org/apache/accumulo/access/grammar/antlr/AccessExpressionAntlrEvaluator.java
```diff
@@ -0,0 +1,121 @@
...
+      a.asSet().stream().forEach(auth -> {
+        e.authorizations.add(auth);
+        String quoted = AccessExpression.quote(auth);
+        if (!quoted.startsWith("\"")) {
+          quoted = '"' + quoted + '"';
+        }
+        e.authorizations.add(quoted);
+      });
+    }
+    // System.out.println("AUTHS: " + entities);
```
keith-turner, line 53:
```suggestion
```

contrib/antlr4/src/test/java/org/apache/accumulo/access/grammar/antlr/AccessExpressionAntlrEvaluator.java
```diff
@@ -0,0 +1,121 @@
...
+        } else if (node instanceof And_expressionContext) {
+          retval = trueCount > 0 && falseCount == 0;
+        } else {
+          retval = trueCount > 0 && falseCount == 0;
+        }
+      }
+      // System.out.println("child results-> true=" + trueCount +", false=" + falseCount);
+    }
+    // System.out.println("node: " + node.getClass().getSimpleName()+ ", value: " + node.getText() +
+    // ", retval: " + retval);
```
keith-turner, line 117: If these were added for debugging and are no longer needed, could remove.  If they may still be useful please ignore this comment.

```suggestion
```

contrib/antlr4/src/test/java/org/apache/accumulo/access/grammar/antlr/Antlr4Tests.java
```diff
@@ -0,0 +1,204 @@
...
+    for (TestDataSet testSet : testData) {
+
+      List<Authorizations> authSets =
+          Stream.of(testSet.auths).map(Authorizations::of).collect(Collectors.toList());
+      AccessEvaluator evaluator = AccessEvaluator.builder().authorizations(authSets).build();
+      AccessExpressionAntlrEvaluator antlr = new AccessExpressionAntlrEvaluator(authSets);
+
+      for (TestExpressions test : testSet.tests) {
+        for (String expression : test.expressions) {
+          // System.out.println("Testing: " + expression);
```
keith-turner, line 133:
```suggestion
```

pom.xml
```diff
@@ -13,6 +13,7 @@
   <properties>
     <maven.compiler.source>11</maven.compiler.source>
     <maven.compiler.target>11</maven.compiler.target>
+    <maven.test.redirectTestOutputToFile>true</maven.test.redirectTestOutputToFile>
```
keith-turner, line 16: This is what was making  prints I added to a test go to a file when I running this locally. Was wondering what was doing that.

contrib/antlr4/README.md
```diff
@@ -0,0 +1,21 @@
...
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+-->
+# ANTLR Example
+
+This contrib example contains an [ANTLRv4](https://www.antlr.org/) grammar file (see [AccessExpression.g4](src/main/antlr4/AccessExpression.g4)) that can be used to create AccessExpressions parsers in languages supported by ANTLRv4.
+
+An example [parser](src/test/java/org/apache/accumulo/access/grammar/antlr/AccessExpressionAntlrParser.java) and [evaluator](src/test/java/org/apache/accumulo/access/grammar/antlr/AccessExpressionAntlrEvaluator.java] are used when building this project to confirm that the parsing and evaluation are consistent with the reference Java implementation.
```
keith-turner, line 21: Could add another paragraph about how antlr was investigated as an option for the impl of accumulo access but the performance was not as good as the custom java code.

keith-turner (approved):

PR #5, opened 2023-09-09
removes left recursion from specification

keith-turner:


keith-turner:
> Are these changes based on the the ANTLR4 grammar in https://github.com/apache/accumulo-access/pull/4/files#diff-f1fd96e8a466678a24201ba1c54b96d5a1d0024e3f734d3eebac3aaae7ddfcf3? Just wondering if I need to make changes in that based on these changes.

Not quite, that still has left recusion with rules `expression : expression AND_OPERATOR expression`.  This PR is an attempt to remove all left recursion while still parsing the exact same language.  So I think it would still reject something like `A|B&C`.  I think that ANTLR grammar would accept `A|B&C`.  I am curious if this set of rules w/o left recursion could almost directly be adapted to ANTLR.

W/o left recursion each rule has a character or set of characters that starts the rule and makes choosing the next rule unnambiguous based on the next character/token seen.  I think with this simplified grammar a parser could be written with a tokenizer and a function for each rule, but not sure.  Thinking this set of rules more easily converts directly to code for parsing because the lack of ambiguity about which rule to follow next makes it easy to code.  So maybe this version of the grammar is easy to code directly or use with something like ANTLR or Javacc.

keith-turner:
I tried literally converting the new spec in this PR to code in this [branch][1] creating [Parser.java][2] and [ParserTest.java][3] and it worked out really nicely,  Passes all the test we have for parsing expressions.  We could use this as a concrete example for writing a parser and also use the concepts in it as a basis for cleaning up the current java code to make it align more closely to the spec.

[1]: https://github.com/keith-turner/accumulo-access/tree/new-parser
[2]: https://github.com/keith-turner/accumulo-access/blob/new-parser/src/main/java/org/apache/accumulo/access/Parser.java
[3]: https://github.com/keith-turner/accumulo-access/blob/new-parser/src/test/java/org/apache/accumulo/access/ParserTest.java

keith-turner:
After comparing the code to the spec I realized I could simplify the spec and did that in 373c192

dlmarion (approved):
I updated the ANTLR grammar in #4 based on these changes here and was able to get that working too.

ctubbsii (approved):
These changes look sensible at a glance. Would need to spend more time looking at the full grammar to be sure there's no issues with it, though. I would go ahead and merge this, with the understanding that reviewing the grammar would be a useful pre-release step to revisit.

PR #6, opened 2023-09-15
adds explantion of evaluation to specification

keith-turner:


ctubbsii (changes requested):
Overall, looks great. I love the boolean algebra example, and the citations to the wiki. I would suggest a few style things:

1. Use consistent wrapping throughout the document. Some of these lines are very long. Maybe wrap around 72 chars like an email message, or 80? Or maybe 100? Just something reasonable, so we don't have some lines wrapping and other lines going off the page forever, requiring a lot of horizontal scrolling to read in a text editor.
2. Use consistent spacing between sentences. We only need 1 space after each period, before the next sentence. Any others are redundant and collapsed in the rendered HTML, so they are pointless, and it can be confusing when doing "find" in a page, and matching a single space in HTML, but two spaces in the Markdown. Some of the sentences have 1 space, others have 2. I even saw one sentence with 3 spaces after the period.
3. Instead of using a mix of inline Markdown links `[link text](link)` and footnote style Markdown links `[link text][footnoteref]`, just use the footnote style. This keeps the document more readable in plain text when viewing the Markdown, and doesn't affect the rendered HTML.

SPECIFICATION.md
```diff
@@ -56,5 +56,41 @@ The definition of utf8 was borrowed from this [ietf document][2].  TODO that doc
...
+
+Evaluation of access expressions performs a combination of [set](https://en.wikipedia.org/wiki/Set_(mathematics)) existence checks and [boolean algebra](https://en.wikipedia.org/wiki/Boolean_algebra). Access expression use the following from boolean algebra.
+
+ * The symbol `&` in an access expression represents [logical conjunction](https://en.wikipedia.org/wiki/Logical_conjunction) which is represented in a boolean algebra as `∧`.
+ * The symbol `|` in an access expression represents [logical disjunction](https://en.wikipedia.org/wiki/Logical_disjunction) which is represented in a boolean algebra as `∨`.
+
+When evaluating an access expression set existence checks are done against a subjects set of authorizations. The following is an algorithm for evaluation an access expression.
+
+ 1. For each access-token in an access expression check if it exists in the subjects set of authorizations.  Replace the access-token with `true` if it exists in the set and `false` otherwise.
+ 2. Evaluate the expression using boolean algebra and only if its true can the subject can access the data labeled with the access expression.
```
dlmarion, line 69:
```suggestion
 2. Evaluate the expression using boolean algebra and only if its true can the subject access the data labeled with the access expression.
```

SPECIFICATION.md
```diff
@@ -56,5 +56,41 @@ The definition of utf8 was borrowed from this [ietf document][2].  TODO that doc
...
+Since `true  ∧ ( false ∨ true )` is true then the subject with authorizations `{RED,GREEN}` can access data labeled with the access expression `RED&(BLUE|GREEN)`.  The access expression `(RED&BLUE)|(GREEN&PINK)` is an example of an access expression that is false for a subject with authorizations `{RED,GREEN}` and it would look like the following using boolean algebra.
+
+ 1. ( RED ∈ {RED,GREEN} ∧ BLUE ∈ {RED,GREEN} ) ∨ ( GREEN ∈ {RED,GREEN} ∧ PINK ∈ {RED,GREEN} )
+ 2. ( true ∧ false ) ∨ ( true ∧ false )
+
+An empty access expression always evaluates to true and this is only thing a subject with the empty set of authorizations can access.
+
+## Escaping
+
+Access tokens can only contain alpha numeric characters or the characters `_`,`-`,`.`,`:`, or `/` unless quoted using `"`.  Within quotes the characters `"` and `\ `' must escaped by prefixing with `\ `.   For example to use `abc\xyz` as an access-token it would need to be quoted and escaped like `"abc\\xyz"`.  When checking if an access-token exist in the subjects authorizations set it must be unquoted and unescaped.
```
dlmarion, line 85:
```suggestion
Access tokens can only contain alpha numeric characters or the characters `_`,`-`,`.`,`:`, or `/` unless quoted using `"`.  Within quotes the characters `"` and `\ `' must escaped by prefixing with `\`.   For example to use `abc\xyz` as an access-token it would need to be quoted and escaped like `"abc\\xyz"`.  When checking if an access-token exists in the subjects authorizations set it must be unquoted and unescaped.
```

SPECIFICATION.md
```diff
@@ -56,5 +56,41 @@ The definition of utf8 was borrowed from this [ietf document][2].  TODO that doc
...
+
+Evaluating `"abc!12"&"abc\\xyz"&GHI`for a subject with authorizations `{abc\xyz,abc!12}` looks like the following in boolean algebra which evaluates to `false`.
+
+ 1. abc!12 ∈ {abc\xyz,abc!12} ∧ abc\xyz ∈ {abc\xyz,abc!12} ∧ GHI ∈ {abc\xyz,abc!12}
+ 2. true ∧ true ∧ false
+
+Notice above when checking if `"abc\\xyz"` exist in the set that it is unquoted and the `\ ` character is unescaped.
```
dlmarion, line 92:
```suggestion
Notice above when checking if `"abc\\xyz"` exist in the set that it is unquoted and the `\` character is unescaped. 
```

dlmarion (approved):

PR #10, opened 2023-09-18
Adds a JMH Benchmark

keith-turner:


src/test/java/org/apache/accumulo/access/AccessExpressionBenchmark.java
```diff
@@ -0,0 +1,183 @@
...
+import java.util.stream.Stream;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+/**
+ * Benchmarks Access Expressions using JMH.  To run, use the following commands.
+ *
+ * <p><blockquote><pre>
+ * mvn clean package
+ * mvn exec:exec -Dexec.executable="java" -Dexec.classpathScope=test -Dexec.args="-classpath %classpath org.apache.accumulo.access.AccessExpressionBenchmark"
```
dlmarion, line 46: Have you tried running this with the `taskset` command to set the Benchmark CPU affinity? This might help remove any OS scheduling from impacting the times. Example command would be:
```
taskset -c 1 mvn exec:exec -Dexec.executable="java" -Dexec.classpathScope=test -Dexec.args="-classpath %classpath org.apache.accumulo.access.AccessExpressionBenchmark"
```

src/test/java/org/apache/accumulo/access/AccessExpressionBenchmark.java
```diff
@@ -0,0 +1,183 @@
...
+
+        int numExpressions =state.getBytesExpressions().size();
+
+        Options opt = new OptionsBuilder()
+                .include(AccessExpressionBenchmark.class.getSimpleName())
+                .operationsPerInvocation(numExpressions)
+                .timeUnit(TimeUnit.MICROSECONDS)
+                .warmupTime(TimeValue.seconds(5))
+                .warmupIterations(3)
+                .measurementIterations(4)
```
dlmarion, line 176: I don't know if there are enough iterations here to trigger the JIT compiler. In a long running server process the `AccessExpression.of` and/or underlying methods will likely be compiled. If you want to simulate that you could either lower the CompilerThreshold or increase the iterations.

dlmarion: [This](https://stackoverflow.com/questions/72482511/ensuring-jmh-benchmark-method-gets-compiled) talks about using the `jvmArgsAppend` parameter of the `Fork` annotation to modify the CompilerThreshold for the benchmark method to 5% of the default (10,000). Another option, maybe in conjunction with lowering the CompilerThreshold, would be iterate over the expressions N times in the test methods, getting over the 10,000 invocation default.

src/test/java/org/apache/accumulo/access/AccessExpressionBenchmark.java
keith-turner: I experimented with the taskset command and it seems like it makes the times a bit better.

src/test/java/org/apache/accumulo/access/AccessExpressionBenchmark.java
keith-turner: I feel like the JIT is kicking in during the runs because it spends 15 seconds in warmup running the methods as many times as it can.  So I think it runs the methods millions of times during warmup.  I don't see things change when giving more time to warmup

dlmarion: Ok, I wasn't sure how many calls had been made.

keith-turner: Looking at [this](https://stackoverflow.com/questions/1503479/how-to-see-jit-compiled-code-in-jvm) I found I could add the following the benchmark builder and see when it compiles methods.  It produces a lot of output.

```
.jvmArgs("-XX:+UnlockDiagnosticVMOptions","-XX:CompileCommand=print,*access/*.*")
```

src/test/java/org/apache/accumulo/access/AccessExpressionBenchmark.java
dlmarion: Yeah, I would expect the times to go down. I'm not sure how the fork parameter you have set is being used. If the tasks are run in parallel, then you might want to pin it to more cores (e.g. 1-4).

src/test/java/org/apache/accumulo/access/AccessExpressionBenchmark.java
dlmarion: I cloned your branch and ran the Benchmark. I think I understand these settings a little more. Here is what I'm seeing:

1. The benchmark runs each benchmark method in a forked JVM
2. The forked JVM run serially, not in parallel
3. The forked JVM runs the benchmark method 3 times as a warmup, timing out if it takes longer than 5s
4. The forked JVM runs the benchmark method 4 times and captures these measurements 
5. In total, each benchmark method is run 12 times.

dlmarion: The number of operations per invocation isn't correct for all of the tests, for example the evaluations tests the number of operations is (numExpressions * numEvaluationTests). I guess as long as it's the same setup for each test, then it should be sufficient to compare two different implementations against each other. Just need to realize that for the evaluations tests, the timing information in the results may not be correct.

dlmarion: > I feel like the JIT is kicking in during the runs because it spends 15 seconds in warmup running the methods as many times as it can. So I think it runs the methods millions of times during warmup. I don't see things change when giving more time to warmup

I didn't realize that it called the benchmark method continuously until the timeout.

keith-turner: > The forked JVM run serially, not in parallel

@dlmarion  one thing I have noticed a few times when running is that a fork will consistently perform differently. Curious if you observed this.  Wonder if the JIT compiles code differently sometimes.

dlmarion: I only noticed minor variations in the ops/us numbers, usually +/- 1.0. I was running this on my EC2 dev box and shutdown the IDE. Nothing else was running. I also ran with `taskset -c 3`.

dlmarion (approved):

PR #12, opened 2023-09-20
improves and tests error messages

### LINKED ISSUE #8: Add unit test that check error messages
For different invalid access expressions it would be nice to have to unit test that check the error messages in the exceptions.

keith-turner:
Improved error messages and added test to verify error messages. Simplified the parser while working on error messages.  Performance of parser is still similar to previous version.

fixes #8


dlmarion (approved):
LGTM

PR #13, opened 2023-09-22
adds a getting started guide

keith-turner:


contrib/getting-started/README.md
```diff
@@ -0,0 +1,89 @@
...
+mvn install
+cd contrib/getting-started
+# Build the example.  If you change the example java code, run this step again.
+mvn clean package
+# Run the example
+./run.sh
+```
+
+## Example runs
+
```
dlmarion, line 45: Might be good to list the data set. data6 is always empty in the runs below, so we don't know why we aren't seeing it.

keith-turner: Made an update in 4a62ab1

dlmarion (approved):

PR #16, opened 2023-09-28
Updates documentation and removes TODOs

keith-turner:


src/main/java/org/apache/accumulo/access/AccessExpression.java
```diff
@@ -88,7 +88,9 @@ static AccessExpression of(String expression) throws IllegalAccessExpressionExce
    * @param expression is expected to be encoded using UTF-8
    */
   static AccessExpression of(byte[] expression) throws IllegalAccessExpressionException {
-    return new AccessExpressionImpl(expression);
+    byte[] copy = new byte[expression.length];
+    System.arraycopy(expression, 0, copy, 0, expression.length);
+    return new AccessExpressionImpl(copy);
```
ctubbsii, line 93:
```suggestion
    return new AccessExpressionImpl(Arrays.copyOf(expression, expression.length));
```

src/main/java/org/apache/accumulo/access/AccessEvaluator.java
```diff
@@ -117,15 +131,31 @@ interface AuthorizationsBuilder {
...
+    /**
+     * When set to a value greater than zero, the result of evaluating expressions will be
+     * remembered and if the same expression is seen it again the remembered result will be used
+     * instead of reevaluating it. If this method is not called on the builder then no caching is
+     * done. When the same expressions can occur repeatedly caching can greatly increase
+     * performance.
+     *
+     * @param cacheSize the number of expressions evaluations to remember in an LRU cache.
+     */
+    OptionalBuilder cacheSize(int cacheSize);
```
ctubbsii, line 158: Instead of having a built-in internal cache with only an option for a fixed size, I wonder if it'd be better to just have a pluggable cache, so users can provide whatever cache implementation they want, rather than only control the size.

keith-turner: Maybe if it would be best to just drop this functionality. It would be really easy for someone to wrap the evaluator with their own cache.  Could just recommend caching in the class level javadocs.

keith-turner: Opened #17

ctubbsii (approved):

PR #22, opened 2023-10-02
fixes build errors

keith-turner:


.github/workflows/maven.yaml
```diff
@@ -70,7 +70,7 @@ jobs:
         profile:
           - {name: 'unit-tests',    javaver: 11, args: 'verify -PskipQA'}
           - {name: 'qa-checks',     javaver: 11, args: 'verify javadoc:jar -Psec-bugs,errorprone'}
-          - {name: 'jdk17',         javaver: 17, args: 'verify javadoc:jar -Psec-bugs,errorprone'}
```
dlmarion, line 73: was errorprone not working with jdk17?

dlmarion (approved):

.github/workflows/maven.yaml
dlmarion: Also, I wonder if we should add a Java 21 build.

keith-turner: Yeah errorprone was failing with jdk17 and I had no idea what to do to fix it.

src/build/ci/spotbugs-exclude.xml
```diff
@@ -0,0 +1,26 @@
...
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+
+-->
+<FindBugsFilter>
+<Match>
+  <Class name="~.*\.jmh_generated\..*" />
```
ctubbsii, line 24: Escaping regexes across multiple languages and libraries is a nightmare. I would strongly advise using `[.]` instead of `\.` for escaping dot, to avoid having to think about how to escape from each point in the hierarchy, as this regex gets parsed through: XML, then Java String, then the Spotbugs library, then Java Regex syntax. It's just too many layers to think about and ensure correctness. It's just so much more reliable and easier to grok if it's `[.]`.

src/main/java/org/apache/accumulo/access/AeNode.java
```diff
@@ -39,6 +39,21 @@ abstract class AeNode implements Comparable<AeNode> {
 
   abstract int ordinal();
 
+  @Override
+  public boolean equals(Object o) {
+    if (o instanceof AeNode) {
+      return compareTo((AeNode) o) == 0;
+    }
+
+    return false;
```
ctubbsii, line 48: this can be written much more succinctly as:

```java
    return o instanceof AeNode && compareTo((AeNode) o) == 0;
```

src/build/ci/spotbugs-exclude.xml
keith-turner: I copied [the pattern](https://spotbugs.readthedocs.io/en/stable/filter.html#match-bugs-in-specific-parts-of-the-code-base) from the projects documentation.

src/main/java/org/apache/accumulo/access/AeNode.java
keith-turner: yeah that is nice

src/build/ci/spotbugs-exclude.xml
ctubbsii: > I copied [the pattern](https://spotbugs.readthedocs.io/en/stable/filter.html#match-bugs-in-specific-parts-of-the-code-base) from the projects documentation.

Understood. Regardless of where the original regex comes from, I still think it's worth using the character class form, rather than the escaped form, to match on literal dots, for the reasons stated. It's not worth editing just to apply this preference, but I think it's worth considering this preference and the reasons for it, whenever one works with regexes going forward.

PR #23, opened 2023-10-03
Change JDK distribution, add build for JDK 21

dlmarion:


dlmarion:
@ctubbsii -  Should we change the JDK distribution from `adopt` to `temurin`? There is a note at [setup-java](https://github.com/actions/setup-java) that says:

> NOTE: AdoptOpenJDK got moved to Eclipse Temurin and won't be updated anymore. It is highly recommended to migrate workflows from adopt and adopt-openj9, to temurin and semeru respectively, to keep receiving software and security updates. See more details in the [Good-bye AdoptOpenJDK post](https://blog.adoptopenjdk.net/2021/08/goodbye-adoptopenjdk-hello-adoptium/).

.github/workflows/maven.yaml
```diff
@@ -68,17 +68,18 @@ jobs:
     strategy:
       matrix:
         profile:
-          - {name: 'unit-tests',    javaver: 11, args: 'verify -PskipQA'}
-          - {name: 'qa-checks',     javaver: 11, args: 'verify javadoc:jar -Psec-bugs,errorprone'}
-          - {name: 'jdk17',         javaver: 17, args: 'verify'}
+          - {name: 'unit-tests',    distribution: adopt,  javaver: 11, args: 'verify -PskipQA'}
+          - {name: 'qa-checks',     distribution: adopt,  javaver: 11, args: 'verify javadoc:jar -Psec-bugs,errorprone'}
+          - {name: 'jdk17',         distribution: adopt,  javaver: 17, args: '-Dmaven.compiler.source=17 -Dmaven.compiler.target=17 verify'}
+          - {name: 'jdk21',         distribution: oracle, javaver: 21, args: '-Dmaven.compiler.source=21 -Dmaven.compiler.target=21 verify'}
```
keith-turner, line 74: The compiler source args are interesting, wonder if that should be done in the accumulo project.

keith-turner (approved):

.github/workflows/maven.yaml
dlmarion: Yeah, I noticed that it was just using the new JDK's to build java 11 compatible code.

keith-turner: That is a good find

ctubbsii: "adopt" in this config just refers to Adoptium's builds, rather than that of the AdoptOpenJDK. So, I'm not sure those articles linked are relevant, as the switch is already accounted for. But, Adoptium doesn't have 21 builds yet, so might as well use Oracle's builds. The best alternative, which I think many people find popular, is the zulu one.

That said, we definitely shouldn't be passing the compiler properties on the command-line. We should be setting all 3 properties, `maven.compiler.source`, `maven.compiler.target`, and `maven.compiler.release` to 11, regardless of which JDK we build with, so those should be in the POM. Really, the only essential one is `maven.compiler.release`, as that replaces the other two for the compiler. But, the reason to still set the other two is that some other Maven plugins (like the formatter) still use those other properties for their own behavior.

ctubbsii (changes requested):
Adding the distribution and JDK21 are fine, but the compiler properties should be set in the POM, along with `maven.compiler.release` to ensure Java 11 compliance... until the project is ready to require something newer.

.github/workflows/maven.yaml
dlmarion: I'll add `maven.compiler.release` to the pom, it's not in there.

dlmarion:
> Adding the distribution and JDK21 are fine, but the compiler properties should be set in the POM, along with `maven.compiler.release` to ensure Java 11 compliance... until the project is ready to require something newer.

Why wouldn't we want to check compliance for other JDK versions? This project is intended to be used outside of Accumulo by other projects, which may be using a different version of the JDK.

.github/workflows/maven.yaml
dlmarion: Added `maven.compiler.release` to pom in 5b3ccb6.

ctubbsii:
> > Adding the distribution and JDK21 are fine, but the compiler properties should be set in the POM, along with `maven.compiler.release` to ensure Java 11 compliance... until the project is ready to require something newer.
> 
> Why wouldn't we want to check compliance for other JDK versions? This project is intended to be used outside of Accumulo by other projects, which may be using a different version of the JDK.

I'm not sure what you're trying to test by adding this. Do you anticipate Java will at some point in the future be changed so it can no longer read or execute Java 11 byte code? The whole point of the `maven.compiler.release` is to ensure that the byte code that is produced will properly and strictly comply with the requirements of a particular Java byte code version, so that future JDK versions can execute it compliant with that older byte code version without suprises. I don't really understand what you're testing if you're trying to build it with different compiler compliance levels.

dlmarion:
> I'm not sure what you're trying to test by adding this.

I was under the impression that compiling with a newer JDK might produce different or more optimized byte code. I did some reading about this last night, and there are only a few cases where this is the case. So, generally, my impression was incorrect. I'll remove the compiler flags from the flags from the command line args.

PR #24, opened 2023-10-04
Bump spotbugs checks to rank 20

ctubbsii:
Bump spotbugs aggressiveness to rank 20 and fix 3 new issues:
* Add missing serialVersionUID field for anonymous subclass of serializable class
* Make use of description field from test data to avoid spotbugs error about unused field (the alternative was to suppress it; I thought printing it was better, since the description is there for our convenience)
* Internal mutable type leaking through benchmark class's API

Also include trivial fixes from code review in #22:
* Simplified .equals to one line
* Use character class for matching on literal dot to avoid escape confusion with backslash

### DIFF
--- a/src/test/java/org/apache/accumulo/access/AccessExpressionBenchmark.java
+++ b/src/test/java/org/apache/accumulo/access/AccessExpressionBenchmark.java
@@ -118,8 +118,8 @@ public class AccessExpressionBenchmark {
       return allTestExpressionsStr;
     }
 
-    public List<EvaluatorTests> getEvaluatorTests() {
-      return Collections.unmodifiableList(evaluatorTests);
+    List<EvaluatorTests> getEvaluatorTests() {
+      return evaluatorTests;
     }
 
   }
--- a/src/test/java/org/apache/accumulo/access/AccessExpressionBenchmark.java
+++ b/src/test/java/org/apache/accumulo/access/AccessExpressionBenchmark.java
@@ -22,7 +22,6 @@ import static java.nio.charset.StandardCharsets.UTF_8;
 
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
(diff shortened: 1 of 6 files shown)


src/test/java/org/apache/accumulo/access/AccessExpressionBenchmark.java
```diff
@@ -117,8 +117,8 @@ List<String> getStringExpressions() {
       return allTestExpressionsStr;
     }
 
-    public ArrayList<EvaluatorTests> getEvaluatorTests() {
-      return evaluatorTests;
+    public List<EvaluatorTests> getEvaluatorTests() {
+      return List.copyOf(evaluatorTests);
```
keith-turner, line 121: This copy will be done on each run of the benchmark that call this function, so it would be good to avoid it.

ctubbsii: I switched to Collections.unmodifiableList instead. That avoids the copy, but keeps spotbugs happy because the encapsulated field is still only exposed in an immutable way.

dlmarion (approved):

src/test/java/org/apache/accumulo/access/AccessExpressionBenchmark.java
```diff
@@ -117,8 +118,8 @@ List<String> getStringExpressions() {
       return allTestExpressionsStr;
     }
 
-    public ArrayList<EvaluatorTests> getEvaluatorTests() {
-      return evaluatorTests;
+    public List<EvaluatorTests> getEvaluatorTests() {
+      return Collections.unmodifiableList(evaluatorTests);
     }
```
keith-turner, line 123: I experimented with this locally. I was curious why it complained about this method that returned a list and not others that returned a list.  It was because the method was public, but it did not need to be public.  So the following change avoids the spotbugs complaint w/o introducing any runtime overhead.

```suggestion
    List<EvaluatorTests> getEvaluatorTests() {
      return evaluatorTests;
    }
```

ctubbsii: Ah, okay. I didn't realize that it didn't need to be public. Thanks!

keith-turner (approved):

PR #25, opened 2023-10-13
Remove caching functionality

### LINKED ISSUE #17: Remove caching functionality
The visibility evaluator builder has an option for caching evaluation execution.  The cache it offers may not satisfy users needs.  It would be very easy for a user to wrap the evaluator with something like a guava cache and use its extensive options.  Therefore it probably not something this project needs to offer.

When removing the cache we could update the class level javadocs to recommend caching for the case where expressions are expected to repeat.

DomGarguilo:
Closes #17 

This PR removes caching functionality and updates the javadoc to reflect its removal.


src/test/java/org/apache/accumulo/access/AccessEvaluatorTest.java
```diff
@@ -89,11 +89,10 @@ public void runTestCases() throws IOException {
         evaluator = AccessEvaluator.builder().authorizations(testSet.auths[0]).build();
         runTestCases(testSet, evaluator);
 
-        evaluator = AccessEvaluator.builder().authorizations(testSet.auths[0]).cacheSize(1).build();
+        evaluator = AccessEvaluator.builder().authorizations(testSet.auths[0]).build();
```
keith-turner, line 92: These two tests runs can be removed because they are now repeating the same tests.

DomGarguilo: Nice catch. Fixed in ea431d5

src/main/java/org/apache/accumulo/access/AccessEvaluator.java
```diff
@@ -24,8 +24,13 @@
  * <p>
  * Used to decide if an entity with one more sets of authorizations can access zero or more access
  * expression.
- *
  * <p>
+ * <p>
+ * Note: For performance improvements, especially in cases where expressions are expected to repeat,
+ * it's recommended to wrap this evaluator with an external caching mechanism, such as Guava's
+ * cache, to leverage its extensive caching options.
```
keith-turner, line 31: Could add a caveat to the caching suggestion.

```suggestion
 * cache, to leverage its extensive caching options.  Caching is only safe under the assumption that for an AccessEvaluator instance that evaluating the same expression multiple times will always yield the same result. When considering caching, may need to mitigate any environmental factors that could change this assumption.
```

keith-turner (approved):

PR #27, opened 2023-10-25
Miscellaneous Improvements

DomGarguilo:


src/main/java/org/apache/accumulo/access/AccessEvaluatorImpl.java
```diff
@@ -121,38 +124,32 @@ static byte[] escape(byte[] auth, boolean quote) {
...
     } else {
       return canAccess(expression.getExpression());
     }
   }
 
   public boolean evaluate(AccessExpressionImpl accessExpression)
       throws IllegalAccessExpressionException {
-    // The VisibilityEvaluator computes a trie from the given Authorizations, that ColumnVisibility
-    // expressions can be evaluated against.
+    // The AccessEvaluator computes a tree from the given Authorizations, that AccessExpressions can
```
dlmarion, line 147: `trie` was not a mis-spelling.

src/main/java/org/apache/accumulo/access/AccessExpressionImpl.java
```diff
@@ -60,7 +55,7 @@ public Authorizations getAuthorizations() {
   }
 
   /**
-   * Creates a column visibility for a Mutation.
+   * Creates an AccessExpression for a Mutation.
```
dlmarion, line 58:
```suggestion
   * Creates an AccessExpression.
```

I don't know that we need to reference Mutation anywhere.

src/main/java/org/apache/accumulo/access/AccessExpressionImpl.java
```diff
@@ -34,11 +34,6 @@ class AccessExpressionImpl implements AccessExpression {
 
   @Override
   public String getExpression() {
-    var expStr = expressionString.get();
-    if (expStr != null) {
-      return expStr;
-    }
-
     return expressionString.updateAndGet(es -> es == null ? new String(expression, UTF_8) : es);
```
dlmarion, line 37: Won't this re-set `expressionString` every time `getExpression()` is called?

DomGarguilo: Yea I think it will. Nice catch

DomGarguilo: I guess both checks for null are needed in case the value changes between the first and second readings of `expressionString`

src/main/java/org/apache/accumulo/access/AccessExpressionImpl.java
```diff
@@ -71,9 +71,9 @@ public Authorizations getAuthorizations() {
   }
 
   /**
-   * Creates a column visibility for a Mutation from a string already encoded in UTF-8 bytes.
+   * Creates an AccessExpression for a Mutation from a string already encoded in UTF-8 bytes.
```
dlmarion, line 74: Mutation exists here too. Might want to search for it, Column Visibility, etc. in the project and fix all of them.

keith-turner (approved):

src/main/java/org/apache/accumulo/access/AccessEvaluatorImpl.java
```diff
@@ -230,4 +226,8 @@ public AccessEvaluator build() {
   public static AuthorizationsBuilder builder() {
     return new BuilderImpl();
   }
+
+  private static boolean isQuoteOrSlash(byte b) {
```
dlmarion, line 230: I think this implementation might perform better. For each invocation of the current method, the input argument `b` has to go through a conversion from `byte` to `char` as described [here](https://docs.oracle.com/javase/specs/jls/se11/html/jls-5.html#jls-5.1.4).

```
    private static final byte QUOTE = (byte) '"';
    private static final byte BACKSLASH = (byte) '\\';

    private static boolean isQuoteOrSlash(byte b) {
      return b == quote || b == backslash;
    }
```

dlmarion: Do you mean in places like:
```
          escapedAuth[index++] = '\\'
```

If so, then I think `yes` we should as `escapedAuth` is a `byte[]`, not a `char[]`. It will save us time on the conversion.

DomGarguilo: Take a look at the additions made in 096df65. Created a new class to house these constants and compare them.

src/main/java/org/apache/accumulo/access/AeNode.java
```diff
@@ -200,7 +203,7 @@ private AndNode(List<AeNode> children) {
 
     @Override
     char operator() {
-      return '&';
+      return AND_OPERATOR;
```
dlmarion, line 206: This method returns a `char`, so I think we want to leave it unless from where it's called its being cast to a `byte`.

src/main/java/org/apache/accumulo/access/AeNode.java
```diff
@@ -249,7 +252,7 @@ private OrNode(List<AeNode> children) {
 
     @Override
     char operator() {
-      return '|';
+      return OR_OPERATOR;
```
dlmarion, line 255: This method returns a `char`, so I think we want to leave it unless from where it's called its being cast to a `byte`.

src/main/java/org/apache/accumulo/access/AeNode.java
```diff
@@ -300,9 +303,9 @@ static AeNode of(Tokenizer.AuthorizationToken auth) {
 
   static AeNode of(byte operator, List<AeNode> children) {
     switch (operator) {
-      case '&':
+      case AND_OPERATOR:
```
dlmarion, line 306: These are comparing bytes, so they are good.

src/main/java/org/apache/accumulo/access/AccessEvaluatorImpl.java
dlmarion: Comments below

DomGarguilo: Addressed your comments in f89ecd0

dlmarion:
I think this looks good. Have you run all the tests and the benchmark?

dlmarion:
It looks like there might not be instructions for running the benchmark. See the comments in #10 for that and output in #7 for what Keith had achieved previously.

DomGarguilo:
> I think this looks good. Have you run all the tests and the benchmark?
> 
> Edit: if you have not run the benchmark, would be good to run against `main` and this branch.

Tests pass. Running the benchmark now on this branch and will run again against main and post results here.

DomGarguilo:
Benchmark results:

# miscImprovements branch

|Benchmark|                                               Mode  |Cnt|   Score|   Error|   Units|
|-------------|-----------|------------|--------|--------|------|
|AccessExpressionBenchmark.measureBytesParsing   |       thrpt |  12 | **10.805**| ± 0.041 | ops/us|
|AccessExpressionBenchmark.measureEvaluation          |  thrpt  | 12 | **13.791**| ± 0.281|  ops/us|
|AccessExpressionBenchmark.measureEvaluationAndParsing | thrpt  | 12 |  **6.552** |± 0.121 | ops/us|
|AccessExpressionBenchmark.measureStringParsing       |  thrpt   |12   |**9.864**| ± 0.051 | ops/us|

# main branch

|Benchmark|                                               Mode  |Cnt|   Score|   Error|   Units|
|-------------|-----------|------------|--------|--------|------|
|AccessExpressionBenchmark.measureBytesParsing   |       thrpt |  12 | **9.450**| ±  0.170 | ops/us|
|AccessExpressionBenchmark.measureEvaluation          |  thrpt  | 12 | **14.012**| ±  0.107|  ops/us|
|AccessExpressionBenchmark.measureEvaluationAndParsing | thrpt  | 12 |  **6.015** |± 0.117 | ops/us|
|AccessExpressionBenchmark.measureStringParsing       |  thrpt   |12   |**8.571**| ± 0.051 | ops/us|

<details>
<summary>Raw output</summary>
<br>

Misc improvements branch:

Result "org.apache.accumulo.access.AccessExpressionBenchmark.measureStringParsing":
  9.864 ±(99.9%) 0.051 ops/us [Average]
  (min, avg, max) = (9.805, 9.864, 9.939), stdev = 0.040
  CI (99.9%): [9.813, 9.914] (assumes normal distribution)


# Run complete. Total time: 00:11:08

REMEMBER: The numbers below are just data. To gain reusable insights, you need to follow up on
why the numbers are the way they are. Use profilers (see -prof, -lprof), design factorial
experiments, perform baseline and negative tests that provide experimental control, make sure
the benchmarking environment is safe on JVM/OS/HW level, ask for reviews from the domain experts.
Do not assume the numbers tell you what you want them to tell.

AccessExpressionBenchmark.measureBytesParsing          thrpt   12  10.805 ± 0.041  ops/us
AccessExpressionBenchmark.measureEvaluation            thrpt   12  13.791 ± 0.281  ops/us
AccessExpressionBenchmark.measureEvaluationAndParsing  thrpt   12   6.552 ± 0.121  ops/us
AccessExpressionBenchmark.measureStringParsing         thrpt   12   9.864 ± 0.051  ops/us

main branch:

Result "org.apache.accumulo.access.AccessExpressionBenchmark.measureStringParsing":
  8.571 ±(99.9%) 0.051 ops/us [Average]
  (min, avg, max) = (8.514, 8.571, 8.621), stdev = 0.040
  CI (99.9%): [8.521, 8.622] (assumes normal distribution)


# Run complete. Total time: 00:11:08

REMEMBER: The numbers below are just data. To gain reusable insights, you need to follow up on
why the numbers are the way they are. Use profilers (see -prof, -lprof), design factorial
experiments, perform baseline and negative tests that provide experimental control, make sure
the benchmarking environment is safe on JVM/OS/HW level, ask for reviews from the domain experts.
Do not assume the numbers tell you what you want them to tell.

AccessExpressionBenchmark.measureBytesParsing          thrpt   12   9.450 ± 0.170  ops/us
AccessExpressionBenchmark.measureEvaluation            thrpt   12  14.012 ± 0.107  ops/us
AccessExpressionBenchmark.measureEvaluationAndParsing  thrpt   12   6.015 ± 0.117  ops/us
AccessExpressionBenchmark.measureStringParsing         thrpt   12   8.571 ± 0.051  ops/us
</details>

dlmarion:
So it looks like`parsing` is faster, and `evaluation` is slower.

dlmarion:
I was wondering earlier about this change in `AccessEvaluatorImpl`.
```
-       .collect(Collectors.toList());
+      .collect(Collectors.toUnmodifiableList());
```

Looking at the sources ([toList](https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/util/stream/Collectors.java#L277), [toUnmodifiableList](https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/util/stream/Collectors.java#L296)), you will see that `toUnmodifiableList` is similar to `toList`, except after collecting to a List it calls `List.of(list.toArray())`.

Could you back out this one change and run the benchmark on your branch again?

dlmarion:
The behavior difference compared to `Collections.unmodifiableList` is interesting. That method passes the list to a class that just delegates the read methods and throws an error on the others. There is no copying of the data.

DomGarguilo:
> I was wondering earlier about this change in `AccessEvaluatorImpl`.
> 
> ```
> -       .collect(Collectors.toList());
> +      .collect(Collectors.toUnmodifiableList());
> ```
> 
> Looking at the sources ([toList](https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/util/stream/Collectors.java#L277), [toUnmodifiableList](https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/util/stream/Collectors.java#L296)), you will see that `toUnmodifiableList` is similar to `toList`, except after collecting to a List it calls `List.of(list.toArray())`.
> 
> Could you back out this one change and run the benchmark on your branch again?

New numbers after changing from `Collectors.toUnmodifiableList()` to `Collectors.toList()`

AccessExpressionBenchmark.measureBytesParsing          thrpt   12  **10.756** ± 0.254  ops/us
AccessExpressionBenchmark.measureEvaluation            thrpt   12  **13.836** ± 0.111  ops/us
AccessExpressionBenchmark.measureEvaluationAndParsing  thrpt   12   **6.600** ± 0.051  ops/us
AccessExpressionBenchmark.measureStringParsing         thrpt   12   **9.823** ± 0.072  ops/us

DomGarguilo:
I ran the benchmark on 6bfcf86

AccessExpressionBenchmark.measureBytesParsing          thrpt   12   8.956 ± 0.068  ops/us
AccessExpressionBenchmark.measureEvaluation            thrpt   12  13.218 ± 0.165  ops/us
AccessExpressionBenchmark.measureEvaluationAndParsing  thrpt   12   5.689 ± 0.051  ops/us
AccessExpressionBenchmark.measureStringParsing         thrpt   12   8.244 ± 0.151  ops/us

I also ran the benchmark on  but removed the helper functions from BytesUtil and just used inline logic with the constants for example from `!isQuoteOrSlash(expression[index])` back to `!(expression[index] == QUOTE || expression[index] == BACKSLASH)`
with these results:

AccessExpressionBenchmark.measureBytesParsing          thrpt   12   8.683 ± 0.077  ops/us
AccessExpressionBenchmark.measureEvaluation            thrpt   12  13.058 ± 0.088  ops/us
AccessExpressionBenchmark.measureEvaluationAndParsing  thrpt   12   5.709 ± 0.036  ops/us
AccessExpressionBenchmark.measureStringParsing         thrpt   12   8.227 ± 0.063  ops/us

Given these results im not sure that the BytesUtil imports are causing any overhead. I am not sure what is causing the difference in performance for Evaluation.

dlmarion:
The error (+/-) range is smaller for the most part with the new changes. I think that is a good thing. I think you should merge what you have at this point. I'm going to look further at this.

dlmarion (approved):

PR #30, opened 2023-10-29
replaces stream with loop to increase performance

keith-turner:
This commit removes a stream used during expression evaluation. Removing it made a noticable performance difference. Seeing the following numbers in the performance test related to evaluation with this change.

```
AccessExpressionBenchmark.measureEvaluation            thrpt   12  20.230 ± 0.331  ops/us
AccessExpressionBenchmark.measureEvaluationAndParsing  thrpt   12   8.415 ± 0.206  ops/us
```

Running the same test without these changes seeing the following.

```
AccessExpressionBenchmark.measureEvaluation            thrpt   12  14.664 ± 0.937  ops/us
AccessExpressionBenchmark.measureEvaluationAndParsing  thrpt   12   6.988 ± 0.352  ops/us
```


dlmarion:
Reading the description I thought that maybe the stream was complicated. But, it's not. It's actually pretty simple and the `allMatch` function's javadoc says that it's short-circuiting.

dlmarion (approved):
We should evaluate any other places where we use `stream`.

PR #31, opened 2023-10-29
Improves performance by avoiding parse tree creation

keith-turner:
This is an experimental commit that improves the performance of evaluating and parsing an expression at the same time by avoiding creation of a parse tree. Also some other steps were taken to avoid other object allocations.

Seeing good performance with these changes when running the benchmark. Comparing the following with numbers from #30, seeing an increase from 8.415 to 10.153 which is a 20% increase.  The perfomance numbers from #30 were run on the same hardware.

```
AccessExpressionBenchmark.measureEvaluationAndParsing  thrpt   12  10.153 ± 0.147  ops/us
```

Not sure how this change fits in with overall code which is why it a WIP. Need to determine how it fits in with the existing parse tree code. For example could creating an expression only validate it w/o creating a parse tree?  Also this code could be slower than creating a parse tree for the case of having multiple authorizations sets.

Another reason this is a WIP is that it could be improved to short circuit in "and" and "or" expression and switch to a validation mode when the result is known and avoid further set lookups.  This is something that is done in the parse tree and could give a further bump in performance.


keith-turner:
Added shot circuiting in 79dacd4 and seeing good performance from that, however I do not like how its implemented.  Feels hacky and still does an operation that could be avoided.   This is now at a 47% increase in performance over #30.

```
AccessExpressionBenchmark.measureEvaluationAndParsing  thrpt   12  12.331 ± 1.462  ops/us
```

src/main/java/org/apache/accumulo/access/ParserEvaluator.java
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  private static boolean parseExpression(Tokenizer tokenizer,
+      Predicate<BytesWrapper> authorizedPredicate, BytesWrapper lookupWrapper) {
+
+    boolean result =
+        parseParenExpressionOrAuthorization(tokenizer, authorizedPredicate, lookupWrapper);
+
+    if (tokenizer.hasNext()) {
+      var operator = tokenizer.peek();
```
dlmarion, line 70: `tokenizer.peek` returns a `byte`, not a `char`. When doing comparisons, use the public static members from `ByteUtils`, AND_OPERATOR and OR_OPERATOR in this case, to remove the byte to char translation.

src/main/java/org/apache/accumulo/access/ParserEvaluator.java
```diff
@@ -0,0 +1,140 @@
...
+      if (result) {
+        // Once the "or" expression is true, can avoid doing set lookups and only validate the rest
+        // of the expression.
+        authorizedPredicate = TRUE_AUTH_PREDICATE;
+      }
+      tokenizer.advance();
+      var nextResult =
+          parseParenExpressionOrAuthorization(tokenizer, authorizedPredicate, lookupWrapper);
+      result |= nextResult;
+    } while (tokenizer.hasNext() && tokenizer.peek() == '|');
```
dlmarion, line 117: Same comment here about byte to char translation

src/main/java/org/apache/accumulo/access/ParserEvaluator.java
```diff
@@ -0,0 +1,140 @@
...
+      if (!result) {
+        // Once the "and" expression is false, can avoid doing set lookups and only validate the
+        // rest of the expression.
+        authorizedPredicate = FALSE_AUTH_PREDICATE;
+      }
+      tokenizer.advance();
+      var nextResult =
+          parseParenExpressionOrAuthorization(tokenizer, authorizedPredicate, lookupWrapper);
+      result &= nextResult;
+    } while (tokenizer.hasNext() && tokenizer.peek() == '&');
```
dlmarion, line 101: Same comment here about byte to char translation

src/main/java/org/apache/accumulo/access/BytesWrapper.java
```diff
@@ -131,4 +131,14 @@ public int hashCode() {
   public String toString() {
     return new String(data, offset, length, UTF_8);
   }
+
+  public void set(byte[] data, int offset, int length) {
```
dlmarion, line 135: Seems like the constructor should call this to avoid duplication.

keith-turner:
Accumulo uses column visibility in the following way.

 * On the client side for insertion a ColumnVisibility object must be created. Evaluation will never be done with this object.
 * On the server side an iterator will parse and evaluate column visibility byte arrays all at once.

So these changes are ideal for the Accumulo server side code.  For Accumulo client side code, it may be useful to offer the ability to validate an expression w/o parsing.  So maybe the direction this PR could take is the following.

 1. Offer functionality to parse and evaluate string or byte[] access expressions all at once w/o any object allocation.
 2. Offer functionality to validate a string or byte[] access expressions w/o any object allocation.
 3. Offer functionality  to compile an access expression.

This PR currently implements 1 above.  For 2 above that could be implemented in this PR and could be used in the client side Accumulo code to make insertions faster.  For 3 above, this library already offers but it could be presented in a different way.

```java
// static method to validate an access expression w/o creating a parse tree, thows an exception if its not valid
AccessExpression.validate("A|B")
// change the existing static methods named of to compile
vat compiledAccessExpression = AccessExpression.compile("A|B);
```

keith-turner:
The following is the result of running the benchmark on commit 7ea7b4e from this PR.

```
AccessExpressionBenchmark.measureBytesValidation   thrpt   12  29.083 ± 1.099  ops/us
AccessExpressionBenchmark.measureEvaluation        thrpt   12  12.429 ± 0.280  ops/us
AccessExpressionBenchmark.measureStringValidation  thrpt   12  24.336 ± 1.423  ops/us
```

The following is the result of running the benchmark on 46b5b877c3eaf433b122812701f47ae5c65ff0c4 from main.

```
AccessExpressionBenchmark.measureBytesParsing          thrpt   12  12.597 ± 0.257  ops/us
AccessExpressionBenchmark.measureEvaluation            thrpt   12  20.407 ± 0.287  ops/us
AccessExpressionBenchmark.measureEvaluationAndParsing  thrpt   12   8.235 ± 0.068  ops/us
AccessExpressionBenchmark.measureStringParsing         thrpt   12  11.267 ± 0.285  ops/us
```

In main the `measureBytesParsing` is roughly equivalent to `measureBytesValidation` in this PR.  That goes from 12.597 to 29.083 which is a 2.31 times faster.  This will be really nice for creating column visibility objects which only need to validate.

In main `measureEvaluation` has no equivalent in this PR because that was testing using an existing parse tree, there is no longer a parse tree.  This was not a useful benchmark because pre-parsed expression would not be available when streaming through data.

In main `measureEvaluationAndParsing` is equivalent to `measureEvaluation` in this PR.  That goes from  8.235 to 12.429 which is 1.51 times faster.  This will be a nice speedup for evaluating visibility expressions on the server side in Accumulo.

keith-turner:
Update https://github.com/apache/accumulo/pull/3746 to use the updated API in this PR.

src/main/java/org/apache/accumulo/access/AccessExpression.java
```diff
@@ -34,70 +30,62 @@
...
-   * As an example of sorting, the expression {@code (Z&Y)|(C&B)} can be sorted to
-   * {@code (B&C)|(Y&Z)}
-   *
-   * <p>
-   * As an example of deduplication, the expression {@code X&Y&X} is equivalent to {@code X&Y}
-   *
-   * @return A normalized version of the access expression that removes duplicates and orders the
-   *         expression in a consistent way.
-   */
-  String normalize();
```
dlmarion, line 77: I think we should keep the normalize method. Deduplicating, sorting, and flattening is a very useful feature. It may be expensive to do, but for a given Key in Accumulo it's only done once when writing the data but may yield better compression performance if many Keys in a block use the same sorted value.

src/main/java/org/apache/accumulo/access/AccessExpression.java
```diff
@@ -34,70 +30,62 @@
...
+ *     {@code
+ * AccessExpression.validate("A&B|C");
+ * }
+ * </pre>
+ *
+ *
  * @see <a href="https://github.com/apache/accumulo-access">Accumulo Access Documentation</a>
  * @since 1.0.0
  */
 public interface AccessExpression {
```
dlmarion, line 58: I think this Object could use a Builder. The quote and normalize methods return strings, so it's not readily apparent how to create the AccessExpression. If you wanted to quote and normalize, would you need to create multiple AccessExpression objects?

Also, since the quote method will only quote if quoting is necessary, should we quote each term in the access expression if needed automatically? 

If we keep the normalize method as an optional operation, then we could use it in the following manner:
```
AccessExpression expr = AccessExpression.builder().of("my&expression").normalize().build();
```

src/main/java/org/apache/accumulo/access/AccessExpression.java
keith-turner: > It may be expensive to do, but for a given Key in Accumulo it's only done once when writing the data but may yield better compression performance if many Keys in a block use the same sorted value.

That is a good reason to keep the normalize method.  I thought of a way to implement it that may be simpler code wise than what was there before, but not completely sure its workable.  Going to try and see what it looks like.

src/main/java/org/apache/accumulo/access/AccessExpression.java
dlmarion: Thinking about this more - if we moved the automatic quoting into the `normalize` method, then we could have just two ways for creating AccessExpression and maybe not need the builder.

```
AccessExpression.of(String)
AccessExpression.normalize(String)
```

The callers uses `of()` when they want the AccessExpression to use the literal string that they supply - we only do validation. The caller uses `normalize()` when they want deduplication, sorting, quoting, and flattening.

keith-turner: > so it's not readily apparent how to create the AccessExpression

At this point the way things have evolved, nothing else really uses the AccessExpression type in the API.  So it seemed like it made sense to turn the object into a bag of static methods.  Some of the static methods do not operate on an access expression, but on parts of it and are used for constructing an expression.  For the methods that do operate on a whole expression, could use a builder pattern like the following.

```java
AccessExpression.builder().of("my&expression").normalize().build()
// get unique authorizations
AccessExpression.builder().of("my&expression").build().getAuthorizations();
// validate the access expression
AccessExpression.builder().of("my&expression").build().validate();
//the remaining static method would be quote which is used to build an expression
AccessExpression.quote("XYZ");
```

The above seems ok with me.  The only one I have a concern about is the `validate` method which minimizes object allocation in this PR, so creating the access expression object to validate would create another object.  We could possibly still have a static validate method like the following.

```java
AccessExpression.builder().of("my&expression").normalize().build()
// get unique authorizations
AccessExpression.builder().of("my&expression").build().getAuthorizations();
// The two static methods on the type
AccessExpression.validate("my&expression");
AccessExpression.quote("XYZ");
```

dlmarion (commented):
Looks good, just a couple minor comments. Please merge when ready, I don't think I need to look at this again.

src/main/java/org/apache/accumulo/access/AccessExpression.java
```diff
@@ -59,47 +74,103 @@ public interface AccessExpression {
...
-   * <p>
-   * As an example of deduplication, the expression {@code X&Y&X} is equivalent to {@code X&Y}
+   * <ul>
+   * <li>As an example of flattening, the expression {@code A&(B&C)} can be flattened to
+   * {@code A&B&C}.</li>
+   * <li>As an example of sorting, the expression {@code (Z&Y)|(C&B)} can be sorted to
+   * {@code (B&C)|(Y&Z)}</li>
+   * <li>As an example of deduplication, the expression {@code X&Y&X} is equivalent to
+   * {@code X&Y}</li>
+   * <li>As an example of unneed parens, the expression {@code "ABC"&"XYZ"} is equivalent to
```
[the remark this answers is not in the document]
keith-turner: Unneeded quotes and parens will be removed, I'll update the docs for that.

src/main/java/org/apache/accumulo/access/AccessExpressionImpl.java
```diff
@@ -20,99 +20,77 @@
...
-  // must create this after creating EMPTY_NODE
-  static final AccessExpression EMPTY = new AccessExpressionImpl("");
+  AccessExpressionImpl(byte[] expression, boolean normalize) {
+    if (normalize) {
+      // validate and normalize expression
+      this.expression = normalize(expression);
+    } else {
+      validate(expression);
+      this.expression = new String(expression, UTF_8);
+    }
```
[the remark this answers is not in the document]
keith-turner: Did not call the string constructor because there are versions of the normalize and validate method that take byte[]

PR #34, opened 2024-01-25
Some touch-ups before release.

dlmarion:
1. Fixed some javadoc
2. Removed unused code
3. Removed use of streams
4. Removed Builder from AccessEvalutor for consistency
5. Added note in javadoc for API objects about UTF-8 expectations

### DIFF
--- a/src/main/java/org/apache/accumulo/access/AccessEvaluator.java
+++ b/src/main/java/org/apache/accumulo/access/AccessEvaluator.java
@@ -18,12 +18,8 @@
  */
 package org.apache.accumulo.access;
 
-import static java.nio.charset.StandardCharsets.UTF_8;
-
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
-import java.util.Set;
 
 /**
  * <p>
@@ -88,12 +84,7 @@ public interface AccessEvaluator {
    * @return AccessEvaluator object
    */
   static AccessEvaluator of(Authorizations authorizations) {
-    final Set<String> authSet = authorizations.asSet();
-    final List<byte[]> authList = new ArrayList<>(authSet.size());
-    for (final String auth : authSet) {
-      authList.add(auth.getBytes(UTF_8));
-    }
-    return new AccessEvaluatorImpl(authList);
+    return new AccessEvaluatorImpl(AccessEvaluatorImpl.convert(authorizations));
   }
 
   /**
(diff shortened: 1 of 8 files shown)


dlmarion:
```
main:

AccessExpressionBenchmark.measureBytesValidation   thrpt   12  32.489 ± 0.166  ops/us
AccessExpressionBenchmark.measureEvaluation        thrpt   12  12.720 ± 0.961  ops/us
AccessExpressionBenchmark.measureStringValidation  thrpt   12  25.196 ± 0.037  ops/us

this branch:

AccessExpressionBenchmark.measureBytesValidation   thrpt   12  32.560 ± 0.127  ops/us
AccessExpressionBenchmark.measureEvaluation        thrpt   12  13.292 ± 0.423  ops/us
AccessExpressionBenchmark.measureStringValidation  thrpt   12  24.660 ± 1.305  ops/us
```

src/main/java/org/apache/accumulo/access/AccessEvaluatorImpl.java
```diff
@@ -44,17 +43,33 @@ class AccessEvaluatorImpl implements AccessEvaluator {
...
+
+    final List<Predicate<BytesWrapper>> predicates = new ArrayList<>(authorizationSets.size());
+    for (final List<byte[]> authorizationList : authorizationSets) {
+      final Set<BytesWrapper> authBytes = new HashSet<>(authorizationList.size());
+      for (final byte[] authorization : authorizationList) {
+        authBytes.add(new BytesWrapper(AccessEvaluatorImpl.escape(authorization, false)));
+      }
+      predicates.add((auth) -> authBytes.contains(auth));
+    }
+    authorizedPredicates = Collections.unmodifiableList(predicates);
```
keith-turner, line 72: Assuming authorizedPredicates is read a lot more frequently than its created, the following would be a bit more expensive to create but probably faster to read because there is less indirection.

```suggestion
    authorizedPredicates = List.copyOf(predicates);
```

src/main/java/org/apache/accumulo/access/AccessEvaluator.java
```diff
@@ -74,88 +82,112 @@ public interface AccessEvaluator {
...
+   * <td>False</td>
+   * </tr>
+   *
+   * </table>
+   *
+   *
+   *
+   */
+  static AccessEvaluator of(Collection<Authorizations> authorizationSets) {
+    final List<List<byte[]>> authorizationLists = new ArrayList<>(authorizationSets.size());
```
keith-turner, line 161: I think its nice to minimize the code in the interface in order to make it easier for someone to quickly read the entire source code for the interface.  Could push this into a static method in AccessEvaluatorImpl in order to minimize the size of the interface source code.  This is a style comment, so feel free to ignore.

keith-turner (approved):
These changes look good. Need to update the getting started [example](https://github.com/apache/accumulo-access/blob/cb6a2d4171ae7c1068b0cf89343edaa46726fbdf/contrib/getting-started/src/main/java/gse/AccessExample.java#L30) with these API changes.

src/main/java/org/apache/accumulo/access/AccessEvaluator.java
dlmarion: Addressed in 549f9c2

src/main/java/org/apache/accumulo/access/AccessEvaluatorImpl.java
dlmarion: Addressed in 549f9c2

PR #37, opened 2024-01-30
Changes to Accumulo-Access for integration with Accumulo

dlmarion:


src/main/java/org/apache/accumulo/access/AccessEvaluatorImpl.java
```diff
@@ -86,6 +86,9 @@ static Collection<List<byte[]>> convert(Authorizations authorizations) {
   AccessEvaluatorImpl(final Collection<List<byte[]>> authorizationSets) {
 
     for (final List<byte[]> auths : authorizationSets) {
+      if (auths == null) {
```
keith-turner, line 89: why allow nulls here?

dlmarion: There was a test failing in apache/accumulo#4206. I think the issue was in the case when the user does not have any authorizations (and maybe no authorizations on the data).

dlmarion:
@keith-turner - If we merge apache/accumulo#3746, then we can just close this without merging?

ctubbsii:
> @keith-turner - If we merge [apache/accumulo#3746](https://github.com/apache/accumulo/pull/3746), then we can just close this without merging?

In anticipation of merging that, or something like it, I will close this now, so we can have a better idea of what's left open to do.

PR #47, opened 2024-02-13
Removed non-applicable information from LICENSE file

dlmarion:


ddanielr (approved):
I think these entries fall under the "unnecessary information" section of the release policy so removing them makes sense.

https://apache.org/legal/release-policy.html#licensing-documentation 
```
Each package MUST provide a LICENSE file and a NOTICE file which account for the package's exact content. LICENSE and NOTICE MUST NOT provide unnecessary information about materials which are not bundled in the package, such as separately downloaded dependencies.
```

dlmarion:
>I think these entries fall under the "unnecessary information" section of the release policy so removing them makes sense.

That's my take on it as well.

PR #48, opened 2024-02-13
Removed contrib/getting-started/.gitignore, updated contrib/.gitignore

dlmarion:


dlmarion:
contrib/getting-started/.gitignore is missing a license header. Resolved this by removing the file and updating the .gitignore file at a higher level in the directory hierarchy.

contrib/.gitignore
```diff
@@ -31,3 +30,17 @@
 /antlr4/nb-configuration.xml
 /antlr4/.vscode/
 /antlr4/.factorypath
+/getting-started/target/
```
DomGarguilo, line 33: If we wanted to ignore any target directories couldn't we just add `target/` to cover that? This applies to other entries here too. I am also not certain that we do want to cover all directories with these name, just wanted to point this out/ask about it.

dlmarion: I don't know if `target/` will get the subdirectories, or if it needs to be something like `**/target/`.

ddanielr:
You could also remove `contrib/.gitignore` updating the entries in the repo's root .gitignore file to not be directory specific.  This would remove the need for the duplicate entries in `contrib/.gitignore`

```
 # Maven ignores
target/
 
 # IDE ignores
.settings/
.project
.classpath
.pydevproject
.idea
*.iml
*.ipr
*.iws
nbproject/
nbactions.xml
nb-configuration.xml
.vscode/
.factorypath
```

contrib/.gitignore
DomGarguilo: Im pretty sure it will catch subdirectories too. [Found this that says as much](https://stackoverflow.com/questions/1470572/ignoring-any-bin-directory-on-a-git-project).

ddanielr: > I don't know if `target/` will get the subdirectories, or if it needs to be something like `**/target/`.

`target/` will get the subdirectories. 
I just tested that by removing the `contrib/.gitignore` file and running `mvn clean package` in `accumulo-access/contrib/antlr4`.
The `accumulo-access/contrib/antlr4/target` directory showed up under untracked files. 

I then modified the entry in the `accumulo-access/.gitignore` from `/target/` to `target/` and my `accumulo-access/contrib/antlr4/target` directory and contents were successfully ignored.

ctubbsii: If you do /target/, then it will only work at the root.

ddanielr (approved):

PR #50, opened 2024-02-13
rename contrib to examples

keith-turner:


DomGarguilo:
I see two more mentions of the word "contrib" in the project:
https://github.com/apache/accumulo-access/blob/caf21fa28ab3c4486d67d6719565f3466483b0bc/examples/antlr4/README.md?plain=1#L19

and "this contrib project" here:
https://github.com/apache/accumulo-access/blob/caf21fa28ab3c4486d67d6719565f3466483b0bc/examples/antlr4/README.md?plain=1#L25

Do we want to change these or leave them?

ctubbsii (changes requested):
Instead of merely changing the name of the directory, I'd like to try to merge the ANTLR stuff into the main project, either as src/main/resources, or as src/test/resources. For the other example, I'd like to try to figure out a better place for that to live. Ultimately, it'd be nice to have the example on the website in a blog post, but I think it could live in the accumulo-examples repo instead. I don't think it needs a full Maven POM file or a run.sh when a java file can be run like:


```
CLASSPATH=/path/to/accumulo-access-$VERSION.jar java AccessExample.java RED PINK
```

I'm going to work on a separate PR to simplify this.

keith-turner:
>Instead of merely changing the name of the directory, I'd like to try to merge the ANTLR stuff into the main project, either as src/main/resources, or as src/test/resources.

I think this is a nice improvement that could be merged for now.  This changes makes the project a bit better than it was before, so not sure it makes sense to not make the change because of possible future improvements.

keith-turner:
@DomGarguilo  I updated the readme in f73e363

DomGarguilo (approved):

ctubbsii:
> > Instead of merely changing the name of the directory, I'd like to try to merge the ANTLR stuff into the main project, either as src/main/resources, or as src/test/resources.
> 
> I think this is a nice improvement that could be merged for now. This changes makes the project a bit better than it was before, so not sure it makes sense to not make the change because of possible future improvements.

Except that renaming it conflicts with the work I'm currently in process of doing. See PR #51

ctubbsii:
This is superseded by #51 (already merged) and #53

PR #51, opened 2024-02-13
Move and simplify the AccessExample

ctubbsii:
* Put the example in the src/test/java directory
* Remove contrib/getting-started
* Simplify the instructions for running the example and add them to the main README
* Bake in some improvements into the example, so it's a bit more self-descriptive, and simplify some of its internals


README.md
```diff
@@ -41,9 +41,34 @@ are package private and are not part of the public API.
...
+<dependencies>
+  <dependency>
+    <groupId>org.apache.accumulo</groupId>
+    <artifactId>accumulo-access</artifactId>
+    <version>$version</version>
+  </dependency>
+</dependencies>
+```
+
+## Running the Example
```
dlmarion, line 58: This shows how to run it, but does not show how to run it with different auths to get different output. See the "Example runs" section of the old README.

keith-turner, line 58: It would be nice to link to the example in the readme so its easy to jump to when browsing on GH.  Not sure how linking in a markdown header looks though.

```suggestion
## Running the [Example](src/test/java/example/AccessExample.java)
```

ctubbsii: I converted that into behavior in the example code itself. If you run it this way, it tells you that you can provide the auths to see a specific example, and runs it through the examples from the old README. I can tweak the message, or add an extra line to the README to make it more clear. Which would be better?

README.md
```diff
@@ -41,9 +41,34 @@ are package private and are not part of the public API.
...
+
+```
+mvn clean package
+CLASSPATH=target/accumulo-access-$version.jar java src/test/java/example/AccessExample.java
```
keith-turner, line 62: This makes it easier to run the command as is.

```suggestion
CLASSPATH=$(ls target/accumulo-access-*.jar) java src/test/java/example/AccessExample.java
```

README.md
ctubbsii: Can try it

ctubbsii: I did some clarification in 48e7c2836ab4560eb449f5f1e5269bfb6961924b

keith-turner:
One nice thing about this structure is that it makes it possible to test the example and ensure changes to the library do not break the example and visa versa.  Experimented with this locally and came up with the following.

```java
package example;

import org.junit.jupiter.api.Assertions;
import org.junit.jupiter.api.Test;

import java.io.ByteArrayOutputStream;
import java.io.PrintStream;
import java.util.List;

public class AccessExampleTest {
    @Test
    public void testExampleCode(){
        final ByteArrayOutputStream baos = new ByteArrayOutputStream();
        final PrintStream out = new PrintStream(baos);
        System.setOut(out);
        AccessExample.main(new String[]{"RED", "BLUE"});
        var output = baos.toString();
        for(var expected : List.of("data3","data5","data6","data9")) {
            Assertions.assertTrue(output.contains(expected));
        }
        for(var unexpected : List.of("data1","data4","data7","data8")) {
            Assertions.assertFalse(output.contains(unexpected));
        }


    }
}

```

README.md
```diff
@@ -41,9 +41,34 @@ are package private and are not part of the public API.
...
+    <version>$version</version>
+  </dependency>
+</dependencies>
+```
+
+## Running the [Example](src/test/java/example/AccessExample.java)
+
+```
+mvn clean package
+CLASSPATH=$(ls target/accumulo-access-*.jar) java src/test/java/example/AccessExample.java
```
keith-turner, line 62: I think the following would be nice based on the earlier comment.  This is in addition to clarifying spaces are needed as was done in an earlier commit.  Feel free to ignore this comment if you don't agree.

```suggestion
CLASSPATH=$(ls target/accumulo-access-*.jar) java src/test/java/example/AccessExample.java
CLASSPATH=$(ls target/accumulo-access-*.jar) java src/test/java/example/AccessExample.java RED BLUE
```

ctubbsii:
I added in your test, @keith-turner , as well as a few other tweaks to the POM that were missing.

pom.xml
```diff
@@ -92,9 +92,18 @@
     <maven.compiler.source>11</maven.compiler.source>
     <maven.compiler.target>11</maven.compiler.target>
     <maven.javadoc.failOnWarnings>true</maven.javadoc.failOnWarnings>
+    <maven.site.deploy.skip>true</maven.site.deploy.skip>
+    <maven.site.skip>true</maven.site.skip>
+    <!-- surefire/failsafe plugin option -->
     <maven.test.redirectTestOutputToFile>true</maven.test.redirectTestOutputToFile>
+    <!-- versions-maven-plugin ignore patterns for snapshots, alpha, beta, milestones, and release candidates -->
+    <maven.version.ignore>.+-SNAPSHOT,(?i).*(alpha|beta)[0-9.-]*,(?i).*[.-](m|rc)[0-9]+</maven.version.ignore>
```
keith-turner, line 100: What does this setting do for us?

keith-turner (approved):

pom.xml
ctubbsii: That's for when you run the versions-maven-plugin. It's optional, but I find it useful to batch update versions like: `mvn versions:display-plugin-updates`

dlmarion (commented):
I reviewed this post-merge and it looks good.

PR #53, opened 2024-02-14
Convert antlr4-example into an integration test

### LINKED ISSUE #43: Contrib project pom's do not get updated on release candidate creation
When a release candidate is made for Accumulo-Access, the poms for contrib/antlr4 and contrib/getting-started do not get updated. We don't create official releases of the contrib artifacts, so I think that it's likely ok that their versions remain SNAPSHOTs. However, we might want to create a post-release action to update the version of accumulo-access that they rely on to use the newly released version.

ctubbsii:
* Update README to point to the antlr-example integration test's README for an ANTLR example.
* Move the contrib/antlr4 project into an integration test invoked by the maven-invoker-plugin, that depends on this project, so it doesn't have to be kept up-to-date
* Fix LICENSE to include mention that Abnf.g4 is under the BSD-3-Clause license
* Get a fresh unaltered copy of Abnf.g4 from upstream, use POM config to specify package, rather than modify it
* Remove unnecessary build stuff from example POM
* Move the ABNF spec to src/main/resources and add a test to ensure it matches the contents of the SPECIFICATION.md file
* Clean up unused checks in checkstyle config for dependencies or code that doesn't exist in this project
* Fix link in README to getting started section

This fixes #43


src/test/java/org/apache/accumulo/access/AccessExpressionTest.java
```diff
@@ -189,4 +198,30 @@ public void testEqualsHashcode() {
...
+            AccessExpression.class.getResourceAsStream("specification/AccessExpression.abnf");
+        var inputStreamReader = new InputStreamReader(abnfFileStream, UTF_8);
+        var bufferedReader = new BufferedReader(inputStreamReader)) {
+
+      Predicate<String> abnfComment = line -> line.startsWith(";");
+      Predicate<String> beforeFirstLine = abnfComment.or(String::isBlank);
+      specLinesFromAbnfFile = bufferedReader.lines().dropWhile(beforeFirstLine).collect(toList());
+    }
+
+    // grab from the markdown, but make sure to skip the markdown triple ticks
```
dlmarion, line 219: IIRC, the `AccessExpression.abnf` file exists solely so that we can test that it's valid ABNF. We were unable to figure out how to embed `AccessExpression.abnf` into `SPECIFICATION.md`. Given that you were able to just pull the ABNF out of `SPECIFICATION.md`, then I think we could probably get rid of `AccessExpression.abnf` and `SpecificationTest` (or just move this test method to `SpecificationTest`.

This is a great change.

src/it/antlr4-example/src/test/java/org/apache/accumulo/access/grammar/SpecificationTest.java
```diff
@@ -30,23 +30,21 @@
...
 public class SpecificationTest {
 
   @Test
   public void testAbnfSpecificationParses() throws Exception {
 
-    // The test resource specification.abnf is a copy of the ABNF
-    // from SPECIFICATION.md
-
-    InputStream is = Antlr4Tests.class.getResourceAsStream("/specification.abnf");
+    var is = AccessExpression.class.getResourceAsStream("specification/AccessExpression.abnf");
```
dlmarion, line 47: I think we could remove `specification/AccessExpression.abnf` and pull the abnf specification from `SPECIFICATION.md` with the changes that you made in `AccessExpressionTest.testSpecificationDocumentation`.

dlmarion (commented):
I have reviewed this post-merge, looks good to me and I think we could leverage an addition in this PR to remove `AccessExpression.abnf`.

src/it/antlr4-example/src/test/java/org/apache/accumulo/access/grammar/SpecificationTest.java
dlmarion: I created #60  for this.

src/test/java/org/apache/accumulo/access/AccessExpressionTest.java
ctubbsii: I thought it would be useful to leave it in its own file in src/main/resources, so it could be grabbed from the jar by other projects, if they wanted to use it. As the canonical definition, I think it's useful in the jar so it can be used as a dependency.

PR #64, opened 2024-02-20
Replace `Authorizer` with `Predicate<String>`

### LINKED ISSUE #57: Authorizer can be replaced with `Predicate<String>`
In AccessEvaluator, there is a small type called `Authorizer` that allows users to plug in an object that determines whether an entity being evaluated has an authorization or not. This is basically just a `Predicate<String>` and can be replaced by that.

```java
HAS_AUTH_PREDICATE.test("someAuth");
// is basically the same as
HAS_AUTH_AUTHORIZER.isAuthorized("someAuth");
```

The main difference is that Predicates can be much more easily composed using `and` and `or`, and works better with all the Java Functional ecosystem. There is an argument to be had for granting a new convenience name to an equivalent functional object. But, I don't think it adds much here, and even if it did, it could at least extend `Predicate<String>` with a default method, so either way of calling could be used. But, I still think we could probably just get rid of it entirely, since `Predicate<String>` by itself does the job well enough.

DomGarguilo:
Fixes #57


src/main/java/org/apache/accumulo/access/AccessEvaluator.java
```diff
@@ -88,12 +90,12 @@ static AccessEvaluator of(Authorizations authorizations) {
   }
 
   /**
-   * Creates an AccessEvaluator from an Authorizer object
+   * Creates an AccessEvaluator from a Predicate&lt;String&gt;
```
dlmarion, line 93: I wonder if we need a little more verbiage to explain when the predicate should return true vs false. Given that this is part of the public API we may want to explain that the AccessEvaluator implementation is going to call `test` passing in a single Authorization and it should return true if the exact String matches.

src/main/java/org/apache/accumulo/access/AccessEvaluator.java
```diff
@@ -159,13 +161,4 @@ static AccessEvaluator of(String... authorizations) {
     return new AccessEvaluatorImpl(AccessEvaluatorImpl.convert(authorizations));
   }
 
-  /**
-   * An interface that is used to check if an authorization seen in an access expression is
```
keith-turner, line 163: This javadoc should be moved (with rewording) to the method that now takes a predicate.  Need to explain what the purpose of the predicate is, which this javadoc kinda does.

src/main/java/org/apache/accumulo/access/AccessEvaluator.java
DomGarguilo: Added to the javadoc in 4e80d25 to try to address this.

src/main/java/org/apache/accumulo/access/AccessEvaluator.java
DomGarguilo: Added to the javadoc in 4e80d25 to try to address this.

dlmarion:
Looking at the amount of code that has changed in this PR, it's not much at all. I think all we have done here is remove an interface which makes clear based on the name of the method and argument how it's used and replaced it with a generic object that is unclear. Because of that we have to try and explain to the user what they should implement in javadoc based on how it's going to be used. I'm not sold on the fact that we should make this change.

DomGarguilo:
> Looking at the amount of code that has changed in this PR, it's not much at all. I think all we have done here is remove an interface which makes clear based on the name of the method and argument how it's used and replaced it with a generic object that is unclear. Because of that we have to try and explain to the user what they should implement in javadoc based on how it's going to be used. I'm not sold on the fact that we should make this change.

Makes sense to me. I can go ahead and close this PR for now.

The ticket also mentioned:
> could at least extend Predicate<String> with a default method, so either way of calling could be used

Does anyone have an opinion on whether this might be worth it or not? I am fine with just closing this PR and the ticket but just wanted to explore all the ideas mentioned in the ticket.

keith-turner:
If Authorizer is kept, then it should have the `@FunctionalInterface` annotation to signify intent and make sure that intent is not changed in the future. Since Authorizer is currently a functional interface its very easy to assign a predicate to it.

PR #65, opened 2024-02-21
Use sorted set in Authorizations

### LINKED ISSUE #58: Authorizations set view could be sorted
Currently, Authorizations is implemented as an unordered set (`Set.copyOf(authorizations)`). It also has some unnecessary copying (`Authorizations.of(Collection<String>)` results in two separate calls to `Set.copyOf`, one of which is redundant). When calling `asSet` to view the result, the returned set could be in any order.

It would be nice to make some improvements, so that a particular set of strings had a canonical view. This may also have implications for hashCode and equals if the implementation stores them in different orderings, but I'm not certain about that. It's also just nice to have a sorted view of the authorizations when calling `asSet` or `toString`.

DomGarguilo:
Fixes #58

### DIFF
--- a/src/main/java/org/apache/accumulo/access/Authorizations.java
+++ b/src/main/java/org/apache/accumulo/access/Authorizations.java
@@ -18,8 +18,11 @@
  */
 package org.apache.accumulo.access;
 
+import java.util.Arrays;
 import java.util.Collection;
-import java.util.Set;
+import java.util.SortedSet;
+import java.util.TreeSet;
+import java.util.stream.Collectors;
 
 /**
  * An immutable collection of authorization strings.
@@ -33,13 +36,13 @@ import java.util.Set;
  * @since 1.0.0
  */
 public final class Authorizations {
-  private final Set<String> authorizations;
+  private final SortedSet<String> authorizations;
 
-  private Authorizations(Set<String> authorizations) {
-    this.authorizations = Set.copyOf(authorizations);
+  private Authorizations(SortedSet<String> authorizations) {
+    this.authorizations = authorizations;
   }
 
-  public Set<String> asSet() {
+  public SortedSet<String> asSet() {
     return authorizations;
   }
 
@@ -64,11 +67,12 @@ public final class Authorizations {
   }
 
   public static Authorizations of(String... authorizations) {
-    return new Authorizations(Set.of(authorizations));
+    return new Authorizations(
+        Arrays.stream(authorizations).collect(Collectors.toCollection(TreeSet::new)));
   }
 
   public static Authorizations of(Collection<String> authorizations) {
-    return new Authorizations(Set.copyOf(authorizations));
+    return new Authorizations(new TreeSet<>(authorizations));
   }
 
 }
--- a/src/test/java/org/apache/accumulo/access/AccessExpressionTest.java
+++ b/src/test/java/org/apache/accumulo/access/AccessExpressionTest.java
@@ -35,7 +35,6 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.function.Predicate;
-import java.util.stream.Collectors;
 
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.function.Executable;
@@ -61,11 +60,10 @@ public class AccessExpressionTest {
       assertEquals(2, testCase.size());
       var expression = testCase.get(0);
       var expected = testCase.get(1);
-      var actual = AccessExpression.of(expression).getAuthorizations().asSet().stream().sorted()
-          .collect(Collectors.joining(","));
+      var actual = String.join(",", AccessExpression.of(expression).getAuthorizations().asSet());
       assertEquals(expected, actual);
-      actual = AccessExpression.of(expression.getBytes(UTF_8)).getAuthorizations().asSet().stream()
-          .sorted().collect(Collectors.joining(","));
+      actual = String.join(",",
+          AccessExpression.of(expression.getBytes(UTF_8)).getAuthorizations().asSet());
       assertEquals(expected, actual);
     }
 


src/main/java/org/apache/accumulo/access/Authorizations.java
```diff
@@ -33,13 +36,13 @@
...
+  private final SortedSet<String> authorizations;
 
-  private Authorizations(Set<String> authorizations) {
-    this.authorizations = Set.copyOf(authorizations);
+  private Authorizations(SortedSet<String> authorizations) {
+    this.authorizations = authorizations;
   }
 
-  public Set<String> asSet() {
+  public SortedSet<String> asSet() {
```
keith-turner, line 45: This returns a mutable object that will change the internal state of the object when its mutated.  The javadoc states its immutable.

keith-turner (commented):
I made some comments on #58 about the general direction of this.  IMO we should do the following.
  
 * Sort in toString for nicer output
 * Keep Set instead of SortedSet in the impl and API.  Also keep using Set.copyOf.
 * Add a unit test to ensure that what asSet() returns is immutable.  Try to add something and verify an exception is thrown and the source did not change.

DomGarguilo:
> I made some comments on #58 about the general direction of this. IMO we should do the following.
> 
> * Sort in toString for nicer output
> * Keep Set instead of SortedSet in the impl and API.  Also keep using Set.copyOf.
> * Add a unit test to ensure that what asSet() returns is immutable.  Try to add something and verify an exception is thrown and the source did not change.

Addressed these points in 6aee3e1.

src/main/java/org/apache/accumulo/access/Authorizations.java
```diff
@@ -58,9 +59,12 @@ public int hashCode() {
     return authorizations.hashCode();
   }
 
+  /**
+   * @return a String containing the sorted, comma-separated set of authorizations
+   */
   @Override
   public String toString() {
-    return authorizations.toString();
+    return authorizations.stream().sorted().collect(Collectors.joining(","));
```
keith-turner, line 67: The set to toString will put `[` and `]` around data which is nice.   Could possibly add those.

keith-turner (approved):

src/main/java/org/apache/accumulo/access/Authorizations.java
dlmarion: I have no issue with this, but I'm curious on the difference between this and `new TreeSet<>(authorizations).toString()`

DomGarguilo: Using a TreeSet will put a space after each comma and add the brackets. e.g. `[A, B, Y, Z]`

Using the stream with `.joining(",")` will not add the spaces nor the brackets by default. e.g. `A,B,Y,Z` but the spaces and brackets can be added obviously.

The TreeSet code would be more simple I think but also would have the overhead of creating the new TreeSet (which might be negligible in this case).

keith-turner: Creating a treeset seems fine for this case.  Thinking is that when toString is called its for the purpose of showing a person something, so in that case the expense is worthwhile.  Can optimize later if ever needed.

dlmarion: > The TreeSet code would be more simple I think but also would have the overhead of creating the new TreeSet

I think that assumes that the streaming example creates no objects. 

`Collection.stream` calls `StreamSupport.stream`, which creates a new `ReferencePipeline.Head`:
https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/util/stream/StreamSupport.java#L67

Calling `sorted` on the `ReferencePipeline` creates more objects:
https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/util/stream/ReferencePipeline.java#L462

`Collectors.joining` creates two new objects.

https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/util/stream/Collectors.java#L367

DomGarguilo: Made toString use TreeSet in 51e02f5

src/main/java/org/apache/accumulo/access/Authorizations.java
```diff
@@ -58,9 +59,12 @@ public int hashCode() {
     return authorizations.hashCode();
   }
 
+  /**
+   * @return a String containing the sorted, comma-separated set of authorizations
+   */
   @Override
   public String toString() {
-    return authorizations.toString();
+    return new TreeSet<>(authorizations).toString();
```
dlmarion, line 67: I wonder what the penalty is here when the user doesn't care about sorting? Consider the example:
```
  LOG.info("Authorizations: {}", authorizations);
```

I wonder if we should leave `toString()` the way that it was, and add a new method with this implementation called `toSortedString` or something. i'm usually on the fence regarding changes like this because we are baking in a performance penalty 100% of the time for something that we *think* is a good idea. I'd rather let the user choose if they want that penalty.

DomGarguilo: I'm fine with that. Although in that case do you think its worth even creating a new method like `toSortedString` or should we just let the user use `toSet` and sort and create the string on their own?

dlmarion: I'm fine with letting users do it on their own. We should get the issue originator to weigh in on this.

DomGarguilo: @ctubbsii do you have any input on this as the creator of the issue associated with this PR?

ctubbsii: I think now that Authorizations.of() takes a set, we could just use the set that the user passed in, preserving whatever order it already had. If it's an immutable set already, we can just keep a reference to the same object instead of doing an unnecessary copy. If it's not, we can preserve the order when we do the copy (maybe using `LinkedHashSet.addAll()` wrapped with `Collections.unmodifiableSet`?). The only issue with this approach is that `asSet()` won't return something that has the `SortedSet` interface if it originally was a mutable sorted set, like TreeSet, but that might be an acceptable compromise).

Issue #66, opened 2024-03-05
Set.Of will throw Illegal argument Exception 


phrocker:
https://github.com/apache/accumulo-access/blob/2da31d3643916c020b14c9e61fa632543dbc56b7/src/main/java/org/apache/accumulo/access/Authorizations.java#L67

Set.of will throw an IllegalArgumentException as expected, but as a consumer to the API we should avoid the exception and prune duplicates .

Would love thoughts. I think it's reasonable to suggest that the client of the Authorizations API should do this work. 

I'll submit a PR for this trivial change if there is any concurrence from @keith-turner  or @dlmarion  that we can de-dupe in the of helper method in Authorization

dlmarion:
@phrocker - do you know if `Set.copyOf` does the same?

dlmarion:
The other API objects have good javadoc coverage, I wonder if that's what is missing here.

phrocker:
> The other API objects have good javadoc coverage, I wonder if that's what is missing here.

I think that is fair. If I knew this beforehand I would have guarded against it. 

Apache Accumulo leverages a hashset which guards against [this with the linked code](https://github.com/apache/accumulo/blob/main/core/src/main/java/org/apache/accumulo/core/security/Authorizations.java#L134):

```  java
private static Set<ByteSequence> createInternalSet(int size) {
    if (size < 1) {
      return EMPTY_AUTH_SET;
    } else {
      return new HashSet<>(size);
    }
  }

  private static List<byte[]> createInternalList(int size) {
    if (size < 1) {
      return EMPTY_AUTH_LIST;
    } else {
      return new ArrayList<>(size);
    }

  }

  /**
   * Constructs an authorization object from a collection of string authorizations that have each
   * already been encoded as UTF-8 bytes. Warning: This method does not verify that each encoded
   * string is valid UTF-8.
   *
   * @param authorizations collection of authorizations, as strings encoded in UTF-8
   * @throws IllegalArgumentException if authorizations is null
   * @see #Authorizations(String...)
   */
  public Authorizations(Collection<byte[]> authorizations) {
    checkArgument(authorizations != null, "authorizations is null");
    this.auths = createInternalSet(authorizations.size());
    this.authsList = createInternalList(authorizations.size());
    for (byte[] auth : authorizations) {
      auths.add(new ArrayByteSequence(auth));
    }
    checkAuths();
  }
```

dlmarion:
[...]

PR #67, opened 2024-03-06
Fixes issue #66: adds a restriction info object that the UI can pull if it is present

phrocker:
I didn't add additional null checks, which could be added as tests into AuthorizationsTests.

It felt less clean to ingest the singular test into AccessExpressionTest or AccessEvaluatorTest, but happy to get feedback. I'm new to this particular project so this is kind of a feeler.

### DIFF
--- a/src/main/java/org/apache/accumulo/access/Authorizations.java
+++ b/src/main/java/org/apache/accumulo/access/Authorizations.java
@@ -18,6 +18,7 @@
  */
 package org.apache.accumulo.access;
 
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Set;
 
@@ -64,7 +65,8 @@ public final class Authorizations {
   }
 
   public static Authorizations of(String... authorizations) {
-    return new Authorizations(Set.of(authorizations));
+    // Use Set.copyOf to remove duplicates if they exist
+    return new Authorizations(Set.copyOf(Arrays.asList(authorizations)));
   }
 
   public static Authorizations of(Collection<String> authorizations) {
--- a/src/test/java/org/apache/accumulo/access/AuthorizationTest.java
+++ b/src/test/java/org/apache/accumulo/access/AuthorizationTest.java
@@ -1,0 +1,36 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.accumulo.access;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.Set;
+
+import org.junit.jupiter.api.Test;
+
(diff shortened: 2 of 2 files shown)


src/test/java/org/apache/accumulo/access/AuthorizationTest.java
```diff
@@ -0,0 +1,36 @@
...
+import org.junit.jupiter.api.Test;
+
+public class AuthorizationTest {
+
+  @Test
+  public void testDuplicateExpressions() {
+    // Test that we can de-duplicate a list of String expressions
+    Set<String> expected = Set.of("A&B", "A&C");
+    var authorization = Authorizations.of("A&B", "A&C", "A&B", "A&B", "A&C");
+    assertEquals(expected, authorization.asSet());
```
keith-turner, line 34: Would be nice to check the other method has the same behavior for dupes. Think the following should do it.

```suggestion
    assertEquals(expected, authorization.asSet());
    authorization = Authorizations.of(List.of("A&B", "A&C", "A&B", "A&B", "A&C"));
    assertEquals(expected, authorization.asSet());
```

keith-turner (approved):
These changes look good and re the test location, I think creating the new test class makes sense. 

I don't understand the PR description.

src/main/java/org/apache/accumulo/access/Authorizations.java
```diff
@@ -64,7 +65,8 @@ public String toString() {
   }
 
   public static Authorizations of(String... authorizations) {
-    return new Authorizations(Set.of(authorizations));
+    // Use Set.copyOf to remove duplicates if they exist
+    return new Authorizations(Set.copyOf(Arrays.asList(authorizations)));
```
ctubbsii, line 69: I think it would be better to keep the `Set.of()` so it issues an obvious error to the user if they supply duplicates, since duplicates could indicate a user typo or similar error.

src/test/java/org/apache/accumulo/access/AuthorizationTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.junit.jupiter.api.Test;
+
+public class AuthorizationTest {
+
+  @Test
+  public void testDuplicateExpressions() {
+    // Test that we can de-duplicate a list of String expressions
+    Set<String> expected = Set.of("A&B", "A&C");
+    var authorization = Authorizations.of("A&B", "A&C", "A&B", "A&B", "A&C");
```
ctubbsii, line 33: These are very confusing examples of Authorizations, since they include a `&` which makes them look like an AccessExpression (visibility label) rather than a set of Authorizations, which typically don't contain such a character. In theory, authorizations could contain these, but they would need to be escaped.

Also, the name of the method says "testDuplicateExpressions", when you're really testing duplicate authorizations.

ctubbsii (changes requested):
I think there's more value in explicitly throwing an error if a user tries to supply a duplicate authorization, rather than silently remove duplicates. If this needs to be more clear, I would support improvements to the javadoc, or changing the method name to construct it as: `Authorizations.ofSet()` or similar.

keith-turner:
> I think there's more value in explicitly throwing an error if a user tries to supply a duplicate authorization, rather than silently remove duplicates. 

What the behavior should be in this case is very subjective.  My preference would be to dedupe, however I am not opposed to throwing an exception.  One interesting thing is that the behavior is inconsistent between the two public static `of()` methods.  Currently before this change one method will dedupe while the other throws an exception.  It would be nice to make the two methods have consistent behavior, document the behavior for dupes, and test it.  Whatever change is made, getting it consistent will  be nice.

dlmarion:
I think we can just change the API so that the user must provide us with a deduplicated set of authorizations. Then we don't have to worry about it or throw an exception. My suggestion on the issue - https://github.com/apache/accumulo-access/issues/66#issuecomment-1979782116

dlmarion:
Created another possible solution in #68 that will either dedupe or throw an exception based on a new input parameter.

phrocker:
> These changes look good and re the test location, I think creating the new test class makes sense.
> 
> I don't understand the PR description.

Sorry about the description. Copy and pasta error. I'll close this out since @dlmarion submitting a good middle ground.

PR #68, opened 2024-03-11
Modified Authorizations.of to handle duplicate input elements

### LINKED ISSUE #66: Set.Of will throw Illegal argument Exception

dlmarion:
Authorizations.of will either throw an IllegalArgumentException or deduplicate the input parameters depending on the value of a new boolean parameter.

Closes #66

### DIFF
--- a/src/main/java/org/apache/accumulo/access/Authorizations.java
+++ b/src/main/java/org/apache/accumulo/access/Authorizations.java
@@ -18,6 +18,7 @@
  */
 package org.apache.accumulo.access;
 
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Set;
 
@@ -39,6 +40,11 @@ public final class Authorizations {
     this.authorizations = Set.copyOf(authorizations);
   }
 
+  /**
+   * Returns the set of authorization strings in this Authorization object
+   *
+   * @return set of authorization strings
+   */
   public Set<String> asSet() {
     return authorizations;
   }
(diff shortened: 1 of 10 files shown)


phrocker (approved):
I think this is a good middle ground.

keith-turner (approved):
I don't really like the boolean, but these change fix the problem of the inconsistent behavior and document it and test it.  Personally I think we should drop the boolean and pick one consistent behavior (either dedupe or throw an exception) for both methods and document the behavior in javadoc and test it.

dlmarion:
> I don't really like the boolean, but these change fix the problem of the inconsistent behavior and document it and test it. Personally I think we should drop the boolean and pick one consistent behavior (either dedupe or throw an exception) for both methods and document the behavior in javadoc and test it.

I don't like the boolean or throwing the exception, but this is middle-ground. Personally, I think we should just change the API to accept a Set instead of a Collection or array. That removes all ambiguity and places the burden on the user to supply us with a unique set of values without throwing an exception.

keith-turner:
> Personally, I think we should just change the API to accept a Set instead of a Collection or array. That removes all ambiguity and places the burden on the user to supply us with a unique set of values without throwing an exception.

I like a single method that takes a set.  I saw that comment on #67 and meant to respond to, but did not immediately respond and then forgot about it.  There is only one thing I don't like about the  single set method is that it makes writing examples a bit more verbose, but losing that is not a bit deal.  I don't think the varargs `of(String ... auhs)` method has much practical use that I can think of, like I doubt it would ever be used outside of example or unit test code.  So having the single method that takes a set seems like a good overall solution.

keith-turner:
We may want to consider removing [this code](https://github.com/apache/accumulo-access/blob/2da31d3643916c020b14c9e61fa632543dbc56b7/src/main/java/org/apache/accumulo/access/AccessEvaluator.java#L155-L160) also.  I have no idea what its behavior is for dupes.   Its not documented and I did not look at the impl.  This method also does not have much practical use other than making example code less verbose.

src/main/java/org/apache/accumulo/access/Authorizations.java
```diff
@@ -33,12 +32,18 @@
...
   private final Set<String> authorizations;
 
   private Authorizations(Set<String> authorizations) {
     this.authorizations = Set.copyOf(authorizations);
   }
 
+  /**
+   * Returns the set of authorization strings in this Authorization object
+   *
+   * @return set of authorization strings
```
keith-turner, line 45:
```suggestion
   * @return immutable set of authorization strings
```

src/main/java/org/apache/accumulo/access/Authorizations.java
```diff
@@ -63,12 +68,14 @@ public String toString() {
...
-  public static Authorizations of(String... authorizations) {
-    return new Authorizations(Set.of(authorizations));
-  }
-
-  public static Authorizations of(Collection<String> authorizations) {
-    return new Authorizations(Set.copyOf(authorizations));
+  /**
+   * Creates an Authorizations object from the set of input authorization strings.
+   *
+   * @param authorizations list of authorization strings
```
keith-turner, line 74:
```suggestion
   * @param authorizations set of authorization strings
```

keith-turner (approved):

dlmarion:
I ran the benchmark and compared to main the changes were in the noise.

Main:
```

AccessExpressionBenchmark.measureBytesValidation   thrpt   12  33.087 ± 0.139  ops/us
AccessExpressionBenchmark.measureEvaluation        thrpt   12  12.863 ± 0.109  ops/us
AccessExpressionBenchmark.measureStringValidation  thrpt   12  25.500 ± 0.144  ops/us
```
This branch:
```
AccessExpressionBenchmark.measureBytesValidation   thrpt   12  33.022 ± 0.171  ops/us
AccessExpressionBenchmark.measureEvaluation        thrpt   12  12.949 ± 0.355  ops/us
AccessExpressionBenchmark.measureStringValidation  thrpt   12  25.577 ± 0.314  ops/us
```

ctubbsii (commented):
:+1:

PR #71, opened 2024-07-05
Avoids looping when evaluating single set of authorizations

keith-turner:
Simplified the code to avoid looping when there is a single set of authorizations.  This change was made in an attempt to improve performance and it did very slightly.

Before this change saw the following.

```
Benchmark                                     Mode  Cnt   Score   Error   Units
AccessExpressionBenchmark.measureEvaluation  thrpt   12  13.609 ± 0.300  ops/us
```

And after this change saw the following.

```
Benchmark                                     Mode  Cnt   Score   Error   Units
AccessExpressionBenchmark.measureEvaluation  thrpt   12  13.806 ± 0.254  ops/us
```

The performance improvement is slight so maybe not worthwhile.  However this change simplified the evaluator code a good bit, so that is worthwhile.


keith-turner:
Was looking at #69 and noticed how that code mimicked the existing code with a loops like the following for the Accumulo code.  Was thinking the inner loop is usually a singleton list and it would be nice to get rid of the loop when it is a singleton.  That is what this PR does.

```java
for(var expression : expressions) {
   for(var evaluator : evaluators){
       // evaluate expression using evaluator
   }
}
```

src/main/java/org/apache/accumulo/access/AccessEvaluator.java
```diff
@@ -149,7 +153,32 @@ static AccessEvaluator of(Authorizer authorizer) {
    *
    */
   static AccessEvaluator of(Collection<Authorizations> authorizationSets) {
-    return new AccessEvaluatorImpl(AccessEvaluatorImpl.convert(authorizationSets));
+    List<AccessEvaluatorImpl> evaluators = new ArrayList<>(authorizationSets.size());
+    for (Authorizations authorizations : authorizationSets) {
+      evaluators.add(new AccessEvaluatorImpl(authorizations));
+    }
+
+    return new AccessEvaluator() {
```
dlmarion, line 161: If this returns an implementation of AccessEvaluator, do we still need AccessEvaluatorImpl? If so, do we end up having two different implementations?

keith-turner: > If this returns an implementation of AccessEvaluator, do we still need AccessEvaluatorImpl? If so, do we end up having two different implementations?

Yeah there are two impls of the interface now.  Still having  AccessEvaluatorImpl makes it really easy for the multiple auths impl to use the single auths impl, so thinking it simplifies the code.

dlmarion: Ok, I think I understand this PR better now. You have modified `AccessEvaluatorImpl` such that it only operates on a single `Authorizations` object and moved the function of dealing with multiples into this anonymous object. I have no issue with this, but wondering if we should/could move this anonymous class to a concrete class, call it `AccessEvaluatorsImpl`, `MultiAccessEvaluatorImpl`, or something like that.

keith-turner: > but wondering if we should/could move this anonymous class to a concrete class, call it AccessEvaluatorsImpl, MultiAccessEvaluatorImpl, or something like that.

Yeah can do that, it would keep the interface file source code really compact and it makes stacktraces a bit more readable.

dlmarion (approved):

PR #73, opened 2024-07-08
Implement Serializable and Iterable where appropriate

### LINKED ISSUE #72: Evaluate whether implementations should be / can be Serializable
https://github.com/apache/accumulo/blob/main/core/src/main/java/org/apache/accumulo/core/security/Authorizations.java implements Serializable and Iterable<byte[]>. I think AccessExpression could be Serializable as well.

dlmarion:
Closes #72

### DIFF
--- a/src/main/java/org/apache/accumulo/access/AccessExpression.java
+++ b/src/main/java/org/apache/accumulo/access/AccessExpression.java
@@ -18,6 +18,8 @@
  */
 package org.apache.accumulo.access;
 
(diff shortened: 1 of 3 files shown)


keith-turner (approved):

ctubbsii (commented):
I'm questioning whether this should have been done. Java serialization is pretty awful, and we already have methods useful for serialization and reconstruction on these classes. `getExpression()` and `asSet()`, which are already Serializable.

PR #77, opened 2024-07-08
Add javadoc about AccessExpression normalization

dlmarion:


dlmarion:
I want to add a note about normalization in the javadoc, as I don't think it's well defined. However, looking at ParserEvaluator (https://github.com/apache/accumulo-access/blob/main/src/main/java/org/apache/accumulo/access/ParserEvaluator.java#L95-L109), I'm not sure that it is short-circuiting.

src/main/java/org/apache/accumulo/access/AccessExpression.java
```diff
@@ -26,6 +26,13 @@
  * expressions as arguments in code, consider using this type instead of a String. The advantage of
  * passing this type over a String is that its known to be a valid expression.
  *
+ * Normalization removes duplicates, sorts, flattens, and removes unneeded parentheses or quotes in
+ * the expression. Normalization is an optional process that the user can choose to occur when
```
DomGarguilo, line 30:
```suggestion
 * the expression. Normalization is an optional process that the user can choose to apply when
```

src/main/java/org/apache/accumulo/access/AccessExpression.java
```diff
@@ -26,6 +26,13 @@
...
+ * constructing an AccessExpression. The AccessEvaluator has the ability to short-circuit
+ * evaluation, for example when the left hand side of an OR expression is valid, then it won't need
+ * to evaluate the right side. The user may not want to perform normalization if they are
+ * constructing their AccessExpressions to take advantage of this feature.
```
DomGarguilo, line 34: Although it can be inferred, it might make things clearer to explicitly say why a user may not want to normalize. Is it because normalization may rearrange the expression which could disrupt an intention short-circuit?

src/main/java/org/apache/accumulo/access/AccessExpression.java
```diff
@@ -26,6 +26,13 @@
  * expressions as arguments in code, consider using this type instead of a String. The advantage of
  * passing this type over a String is that its known to be a valid expression.
  *
+ * Normalization removes duplicates, sorts, flattens, and removes unneeded parentheses or quotes in
```
DomGarguilo, line 29:
```suggestion
 * <p>
 * Normalization removes duplicates, sorts, flattens, and removes unneeded parentheses or quotes in
```

src/main/java/org/apache/accumulo/access/AccessExpression.java
dlmarion: Yes, that's the implication. You can see it in the tests. https://github.com/apache/accumulo-access/blob/main/src/test/java/org/apache/accumulo/access/AccessExpressionTest.java#L103 is a good example where the tokens get re-ordered. If the OR short-circuting is occurring (my question above), then I think the user would want to craft the AccessExpression such that the tokens that have a higher probability of being seen occur first in the expression.

dlmarion:
> I want to add a note about normalization in the javadoc, as I don't think it's well defined. However, looking at ParserEvaluator (https://github.com/apache/accumulo-access/blob/main/src/main/java/org/apache/accumulo/access/ParserEvaluator.java#L95-L109), I'm not sure that it is short-circuiting.

I think I have resolved this question myself, it looks like the short-circuiting is happening in the OR case. It's not done by stopping processing, but it continues processing to validate the expression but swaps out the test that is used to see if the token is in the auth set to a test that just returns `true`. The `ParserEvaluator.parseAccessExpression` is called from AccessExpressionImpl and AccessEvaluatorImpl. I can see why we need to validate the expression in AccessExpressionImpl. I'm not sure we need to re-validate the expression in AccessEvaluatorImpl.

keith-turner:
> I'm not sure we need to re-validate the expression in AccessEvaluatorImpl.

Un-validated expression can be passed to the AccessEvaluator because it accepts String and `byte[]`, so its good to still validate those. In the case where a AccessExpression is passed in it is doing redundant work for the short circuit case.

keith-turner (approved):

PR #79, opened 2024-07-18
Put benchmark config inside its execution

ctubbsii:
* Configuration outside of the execution will apply to all executions of exec-maven-plugin when the profile is active, even executions that have nothing to do with the benchmark, so the benchmark-specific configuration should be contained inside the execution element
* Remove extra character from README

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,8 +73,9 @@ For an ANTLRv4 example, see antlr-example integration test's
 This project includes a JMH Benchmark. To run it:
 
 ```
-mvn clean verify -Pbenchmark"
+mvn clean verify -Pbenchmark
 ```
+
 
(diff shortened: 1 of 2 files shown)


keith-turner (approved):

dlmarion (approved):

ctubbsii:
For what it's worth, there is an exec-maven-plugin goal called "java" that is supposed to make it easy to run a mainClass without having to pass the `-cp` argument, but it doesn't correctly pass along the current project's classpath in a convenient way when it forks to execute the mainClass. It's frustrating, but calling `exec:exec` this way with the classpath as an argument to the `java` command is the easiest workaround.

PR #82, opened 2025-01-07
Copies some test cases from accumulo code base

keith-turner:


DomGarguilo (approved):
Ran the updated test and it passes

PR #84, opened 2025-02-11
memoizes creating a parse tree

keith-turner:
This change offers a new parse() method on AccessExpression that creates a parse tree if one does not exits.  This change allows decoupling how access expression are created (w/ or w/o an initial parse tree) from using parse trees.

For example code the like the following that wanted to use a parse tree to validate an access expression would have required the ParsedAccessExpression type to be passed to it.  Now it can accept an AccessExpression and call parse() which will do the most efficient thing depending on how the expression was created.

```java

void checkExpression(AccessExpression expression){
    var parsed = expression.parse();
    // use parse tree to validate expression
}

```

### DIFF
--- a/src/main/java/org/apache/accumulo/access/AccessExpression.java
+++ b/src/main/java/org/apache/accumulo/access/AccessExpression.java
@@ -107,6 +107,15 @@ public abstract class AccessExpression implements Serializable {
    */
   public abstract String getExpression();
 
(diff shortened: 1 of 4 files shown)


ddanielr (approved):
Makes sense and I think allowing the users to not have to chose between object types is a great improvement.

PR #85, opened 2025-02-12
Changes exception prefix from Invalid to Illegal

keith-turner:
Renamed InvalidAccessExpressionException to
IllegalAccessExpressionException because it extends IllegalArgumentException.  So now it uses the same terminology as its parent class.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ section](#getting-started) for an example of how to use this java library.
 The following types constitute the public API of this library. All other types
 are package private and are not part of the public API.
 
-  * [InvalidAccessExpressionException](src/main/java/org/apache/accumulo/access/InvalidAccessExpressionException.java).
+  * [IllegalAccessExpressionException](src/main/java/org/apache/accumulo/access/IllegalAccessExpressionException.java).
   * [AccessEvaluator](src/main/java/org/apache/accumulo/access/AccessEvaluator.java).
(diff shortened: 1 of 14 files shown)


ddanielr (approved):
LGTM

ctubbsii:
@keith-turner are you sure this is a good idea? Changing it to begin with "IllegalAccess" makes it seem like it's related to https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/IllegalAccessException.html or https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/IllegalAccessError.html, which is not what this is. In this case, it's only an illegal argument because the access expression is invalid, so I think the previous name still makes sense. At the very least, the word "illegal" seems to imply "disallowed" but without a reason why. "Invalid", on the other hand, makes it clear that it's disallowed because it's not a valid access expression. Given the potential for confusion, and the loss of the clarity about validity/lack thereof, I think it's better to stick with the old name. I'm not sure there's a compelling reason to make the exception match the parent's name. Making it the same prefix just to make it the same isn't really compelling.

keith-turner:
@ctubbsii  those reasons all SGTM for changing the name back.

ctubbsii:
Okay, reverted in 6112180245fb8e1aed41b8327e4d773ddc735aaf

PR #86, opened 2025-02-13
Removed double check locking in AccessExpressionImpl.parse

dlmarion:


keith-turner (approved):

ctubbsii (approved):
I think the double checked locking was working well enough. It would have guaranteed that no cycles were spent computing the parse tree more than once, but at the small cost of synchronizing if there was a collision on initialization. This PR's implementation also works fine. It might compute the parse tree more than once, the first time it is initialized, but guarantees that the first one that finishes will not get clobbered and avoids the synchronization, and might be a little easier to read.

I think it's fine either way.

Issue #88, opened 2025-11-23
Implementation does not match specification


willmurnane:
In the process of developing [a Rust implementation of this specification](https://github.com/willmurnane/access-rs), I implemented [a check that the Rust implementation's behavior matches the Java code](https://github.com/willmurnane/access-rs/pull/3), and was surprised to find that the Java code does not implement what the ABNF says it does. Specifically, [the ABNF says](https://github.com/apache/accumulo-access/blob/main/SPECIFICATION.md) that access tokens include quoted strings composed of characters 0x20 and greater with a few exceptions, but the Java code accepts the characters 0x0 through 0x1f, as demonstrated by running this code:
```java
package test.access;
import org.apache.accumulo.access.AccessEvaluator;
import org.apache.accumulo.access.AccessExpression;
import org.apache.accumulo.access.Authorizations;
import org.apache.accumulo.access.IllegalAccessExpressionException;

public class Main {
    public static void main(String[] args) {
        for (char i = 0; i < 0x25; i++) {
            String s = String.format("\"%c\"", i);
            System.out.printf("Character 0x%02x %s -> ", (int)i, s);
            try {
                if (i == 0x22) {
                    System.out.printf("Double quote is illegal, skipping\n");
                }
                AccessExpression.of(s);
                System.out.printf(" %s", AccessExpression.of(s));
                System.out.printf(" %s", AccessEvaluator.of(Authorizations.of(String.format("%c", i))).canAccess(s));
                if (i < 0x20) {
                    System.out.printf("(expected failure)\n");
                } else {
                    System.out.printf("(expected success)\n");
                }
            } catch (IllegalAccessExpressionException iaee) {
                System.out.printf("%s Failed!\n", s);
            }
        }
    }
}
```

I'm not sure what the intended behavior is here, should the specification or the code be updated to match the other?

keith-turner:
> I'm not sure what the intended behavior is here, should the specification or the code be updated to match the other?

Good find, we do need to fix this one way or another.

We would like access expression to be human readable.  The current code was created to be compatible with the code in Accumulo which accepts any byte between the quotes (and therefore may not be human readable).  The ABNF was created w/ the goal of being human readable, but the impl does not currently match the spec.  Also looking back uncertain if the ABNF definition of UTF8 is correct, we may have copied it from somewhere w/o validating.

We could potentially do the following to work towards achieving this goal.

[...]

PR #89, opened 2025-12-10
Modify internals to use byte[] instead of String

dlmarion:
The current implementation does not allow for non-UTF8 characters to be used. This change alters the internal code to use byte[] instead of String to allow for that in AccessExpressions and Authorizations objecs.

Related to #88

### DIFF
--- a/src/build/ci/find-unapproved-public.sh
+++ b/src/build/ci/find-unapproved-public.sh
@@ -24,7 +24,7 @@ count=$(grep -E "public.*(class|interface|enum|record)" src/main/java/org/apache
   grep -v " class ParsedAccessExpression " |
   grep -v " enum ExpressionType " |
   grep -v " class Authorizations " |
-  grep -v " class Bytes " |
+  grep -v " interface Bytes " |
   grep -c -v " class InvalidAccessExpressionException ")
 
 files=$(grep -E "public.*(class|interface|enum|record)" src/main/java/org/apache/accumulo/access/*.java |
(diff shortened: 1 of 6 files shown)


src/main/java/org/apache/accumulo/access/BytesWrapper.java
```diff
@@ -22,10 +22,12 @@
 import static java.util.Objects.checkFromIndexSize;
 import static java.util.Objects.checkIndex;
 
+import java.io.Serializable;
 import java.util.Arrays;
 
-final class BytesWrapper implements Comparable<BytesWrapper> {
+public final class BytesWrapper implements Comparable<BytesWrapper>, Serializable {
```
keith-turner: If this will be in the public api then could just call it `Bytes`.  It would need a since tag.  Also could add a `byte[] toArray()` method that returns a copy of the array.  The `byteAt()` method should be made public.

dlmarion: Backed out this change in e04689e. Renamed BytesWrapper to BytesImpl and created a Bytes interface.

src/build/ci/find-unapproved-public.sh
```diff
@@ -24,9 +24,20 @@ count=$(grep -E "public.*(class|interface|enum|record)" src/main/java/org/apache
   grep -v " class ParsedAccessExpression " |
   grep -v " enum ExpressionType " |
   grep -v " class Authorizations " |
+  grep -v " class Bytes " |
   grep -c -v " class InvalidAccessExpressionException ")
 
+files=$(grep -E "public.*(class|interface|enum|record)" src/main/java/org/apache/accumulo/access/*.java |
```
dlmarion: There is probably an easier way to get the count of the `files` array, but I didn't look that hard at the `bash` man page.

PR #95, opened 2025-12-22
Modified entry point for AccessEvaluator

### LINKED ISSUE #56: Consider removing AccessEvaluator as an API or improving its entry points
I think the entry point for AccessEvaluator entry point has room for improvement. Currently, the entry point is a static method named `of` that is overloaded. However, it almost exclusively takes an Authorizations object of some form.

Rather than construct an Authorizations object, then construct a separate AccessEvaluator object, it would probably make more sense to have a method that dangles off of Authorizations that evaluates.

For example:

```java
var evaluator = AccessEvaluator.of(Authorizations.of("a", "b"));
// vs.
var evaluator = Authorizations.of("a", "b").evaluator();
```

What I'm suggesting is similar to the way Java regex Patterns can construct a Matcher. You wouldn't do `Matcher m = new Matcher(Pattern.compile("p"), "other");`. Instead, you do `Pattern.compile("p").matcher("other")`. You start with the thing you have, then you derive the related objects from it. In our case, our starting point is the Authorizations.

dlmarion:
Removed the 'of' methods from the AccessEvaluator
interface. Created an interface for Authorizations and renamed the current class to AuthorizationsImpl. Added the following methods to get an AccessEvaluator:

  1. The evaluator() method on an Authorizations instance
  2. The Authorizations.evalutor() static method to get an instance for multiple Authorizations objects.
  3. The Authorizations.using(Authorizor) static method.

Closes #56

### DIFF
--- a/antlr4-example/src/main/java/org/apache/accumulo/access/antlr4/AccessExpressionAntlrEvaluator.java
+++ b/antlr4-example/src/main/java/org/apache/accumulo/access/antlr4/AccessExpressionAntlrEvaluator.java
@@ -26,7 +26,6 @@ import java.util.Set;
 import org.antlr.v4.runtime.tree.ParseTree;
 import org.antlr.v4.runtime.tree.TerminalNode;
 import org.apache.accumulo.access.AccessExpression;
-import org.apache.accumulo.access.Authorizations;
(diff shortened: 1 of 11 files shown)


DomGarguilo:
Does this close #56 once merged?

PR #96, opened 2025-12-23
Modified accumulo access to only work with String and Unicode.

### LINKED ISSUE #88: Implementation does not match specification

keith-turner:
This is still a work in progress, but its close to being complete. Modified the Accumulo Access code to only use String and unicode.  Made the following changes.

 * Removed all methods that take `byte[]` in the API
 * Removed the many static entry points in the API and replaced them w/ a single static entry point.
 * Reworked the internal code to operate on char instead of byte.
 * Limited authorizations to valid unicode characters that are not ISO control characters by default.  This is configurable though.

Realized it would be really nice if users could further limit authorizations to a smaller subset of characters while working on this.  The existing Accumulo Access APIs had a ton of static entry points and there was no one place that configurable behavior could be introduced .  That led to reworking the API to have a single entry point where configuration could  be applied instead of the many entry static points.  Currently this entry point only allows setting a configurable authorization validator, but it makes it easy to support future configuration.  This new single entry point is conceptually similar to how Gson works, it was inspired by that.

Still puzzling about how Accumulo might use this code or if that is possible.  Want to experiment w/ that some.

Have not yet run the benchmarks w/ these code changes.

fixes #88

### DIFF
--- a/core/src/main/resources/org/apache/accumulo/access/specification/AccessExpression.abnf
+++ b/core/src/main/resources/org/apache/accumulo/access/specification/AccessExpression.abnf
@@ -28,9 +28,8 @@ and-expression          =  "&" (access-token / paren-expression) [and-expression
 or-expression           =  "|" (access-token / paren-expression) [or-expression]
 
 access-token            = 1*( ALPHA / DIGIT / "_" / "-" / "." / ":" / slash )
-access-token            =/ DQUOTE 1*(utf8-subset / escaped) DQUOTE
+access-token            =/ DQUOTE 1*(unicode-subset / escaped) DQUOTE
 
-utf8-subset             = %x20-21 / %x23-5B / %x5D-7E / unicode-beyond-ascii ; utf8 minus '"' and '\'
-unicode-beyond-ascii    = %x0080-D7FF / %xE000-10FFFF
+unicode-subset          = %x00-21 / %x23-5B / %x5D-7F / unicode-beyond-ascii ; unicode minus '"' and '\'
 escaped                 = "\" DQUOTE / "\\"
 slash                   = "/"
--- a/core/src/test/java/org/apache/accumulo/access/tests/AccessExpressionTest.java
+++ b/core/src/test/java/org/apache/accumulo/access/tests/AccessExpressionTest.java
@@ -41,7 +41,6 @@ import org.apache.accumulo.access.AccessExpression;
 import org.apache.accumulo.access.AuthorizationValidator;
 import org.apache.accumulo.access.InvalidAccessExpressionException;
 import org.apache.accumulo.access.ParsedAccessExpression;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.function.Executable;
 
@@ -158,7 +157,6 @@ public class AccessExpressionTest {
     assertNotEquals(ae2.hashCode(), ae4.hashCode());
   }
 
-  @Disabled
   @Test
   public void testSpecificationDocumentation() throws IOException, URISyntaxException {
     // verify AccessExpression.abnf matches what is documented in SPECIFICATION.md


keith-turner:
> Do we still need the quoting and unquoting of terms in an AccessExpression if the goal is String and Unicode only?

Yeah still need them, the quoting allows using characters that are reserved for the language itself in auths.  Also allows the language to use new characters in the future that are also used in quoted expressions.

core/src/main/java/org/apache/accumulo/access/AccumuloAccess.java
```diff
@@ -0,0 +1,225 @@
...
+import org.apache.accumulo.access.impl.BuilderImpl;
+
+/**
+ * The entry point into Accumulo Access to create access expressions, expression evaluators, and
+ * authorization sets.
+ *
+ * @see #builder()
+ * @since 1.0
+ */
+public interface AccumuloAccess {
```
ctubbsii: I don't think we need "Accumulo" prefix on this object. While Accumulo is the containing project, I think this class could just be called "Access".

core/src/main/java/org/apache/accumulo/access/AccumuloAccess.java
```diff
@@ -0,0 +1,225 @@
...
+
+  /**
+   * Quickly validates that an access expression is properly formed.
+   *
+   * @param expression a potential access expression that
+   * @throws InvalidAccessExpressionException if the given expression is not valid
+   * @throws InvalidAuthorizationException if the expression contains an invalid authorization
+   * @throws NullPointerException when the argument is null
+   */
+  void validate(String expression)
```
ctubbsii: I'm not sure I like this here. I get the utility of having it to avoid the object allocation if you just want to quickly verify a bunch of expressions, but it feels out of place. This isn't really here to construct or use an object from this library... it's just kind of an extra utility method that allows users to evaluate their own data. But, I think what users really are asking if they were to use this method is "if I were to construct a new AccessExpression from this String, would it succeed?", and I think the only 100% reliable way to do that is to just create a new AccessExpression from this String, which kind of makes this method redundant.

core/src/main/java/org/apache/accumulo/access/AccumuloAccess.java
```diff
@@ -0,0 +1,225 @@
...
+   * </tr>
+   * <tr>
+   * <td>A&amp;D</td>
+   * <td>False</td>
+   * <td>False</td>
+   * </tr>
+   *
+   * </table>
+   */
+  AccessEvaluator newEvaluator(Collection<Authorizations> authorizationSets);
```
ctubbsii: If AccessEvaluator were a FunctionalInterface (which I think it can, and should, be), I think this basically becomes something like:

```java
AccessEvaluator aggregateEvaluator = expression -> authorizationsSets.stream().map(this::newEvaluator).allMatch(e -> e.canAccess(expression));

for (AccessExpression expression : allExpressionsToEvaluate) {
  if (aggregateEvaluator.canAccess(expression)) {
    // do whatever
  }
}
```

I think this is something a user can do themselves (using streams or traditional loops) if they had this use case. I think this is a useful example to document, but I don't think we need it in the main API entry point.

core/src/main/java/org/apache/accumulo/access/AccumuloAccess.java
keith-turner: Made that change in 778c9e483adb83c1e3eefa3cb824456f0b1dd2f6.  I wanted a shorted name for this class but was not sure what name to use.  `Access` seems good to me.

core/src/main/java/org/apache/accumulo/access/Access.java
```diff
@@ -0,0 +1,225 @@
...
+   * tree is already available, then it would likely be faster to use it rather than call this
+   * method.
+   * </p>
+   *
+   * @throws InvalidAccessExpressionException when the expression is not valid.
+   * @throws InvalidAuthorizationException when the expression contains an authorization that is not
+   *         valid
+   * @throws NullPointerException when any argument is null
+   */
+  void findAuthorizations(String expression, Consumer<String> authorizationConsumer)
```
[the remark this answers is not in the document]
keith-turner: It would be nice to move this out of the root level interface, however if that is done then expression would be parsed twice to do the following.  

```java
Access access = ...;
access.newExpression("a|b|c").findAuthorizations(System.out::println);
```

It will parse once to validate when creating the AccessExpression and then again when calling findAuthorizations().

core/src/main/java/org/apache/accumulo/access/Access.java
```diff
@@ -0,0 +1,225 @@
...
+
+  /**
+   * Quickly validates that an access expression is properly formed.
+   *
+   * @param expression a potential access expression that
+   * @throws InvalidAccessExpressionException if the given expression is not valid
+   * @throws InvalidAuthorizationException if the expression contains an invalid authorization
+   * @throws NullPointerException when the argument is null
+   */
+  void validate(String expression)
```
keith-turner: It is the same as calling newExpression, except it does not allocate an object.  So its an optimization.  Could remove it, it can be added later.    Also could rename it to `validateExpression()` which lines up w/ the `newExpression()` method.

core/src/main/java/org/apache/accumulo/access/Access.java
```diff
@@ -0,0 +1,225 @@
...
+   * </tr>
+   * <tr>
+   * <td>A&amp;D</td>
+   * <td>False</td>
+   * <td>False</td>
+   * </tr>
+   *
+   * </table>
+   */
+  AccessEvaluator newEvaluator(Collection<Authorizations> authorizationSets);
```
keith-turner: > If AccessEvaluator were a FunctionalInterface

Could be tricky to achieve that because it could lead to forcing double parsing (once for validation and once of evaluation) depending on how its done.

core/src/main/java/org/apache/accumulo/access/Access.java
keith-turner: Renamed to `validateExpression` in 7a375938e4ea914b9f5c50ac0bd6de47589d2eda.  Plan to evaluate removal based on trying to use the changes in accumulo.  I believe this is used now in main, but not 100% sure.

core/src/main/java/org/apache/accumulo/access/Access.java
```diff
@@ -0,0 +1,225 @@
...
+   *         valid
+   * @throws InvalidAccessExpressionException if the given expression is not valid
+   */
+  ParsedAccessExpression newParsedExpression(String expression)
+      throws InvalidAccessExpressionException, InvalidAuthorizationException;
+
+  /**
+   * @return a pre-allocated empty Authorizations object
+   */
+  Authorizations newAuthorizations();
```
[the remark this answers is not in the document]
keith-turner: Removed in 8331f6313fd6e8aa12cceb3d1b326743ac72316c

core/src/main/java/org/apache/accumulo/access/impl/AuthorizationsImpl.java
```diff
@@ -0,0 +1,72 @@
...
+ * under the License.
+ */
+package org.apache.accumulo.access.impl;
+
+import java.util.Iterator;
+import java.util.Set;
+
+import org.apache.accumulo.access.Authorizations;
+
+public final class AuthorizationsImpl implements Authorizations {
```
dlmarion: I wonder if this could simplify the code? Records are implicitly `final` and it takes care of the `equals`, `hashcode` and `toString` methods.
```
  public record AuthorizationsImpl(Set<String> authorizations) implements Authorizations
```

core/src/main/java/org/apache/accumulo/access/impl/AccessEvaluatorImpl.java
```diff
@@ -30,42 +29,56 @@
...
 
   /**
    * Create an AccessEvaluatorImpl using an Authorizer object
    */
-  public AccessEvaluatorImpl(Authorizer authorizationChecker) {
-    this.authorizedPredicate = auth -> authorizationChecker.isAuthorized(unescape(auth));
+  public AccessEvaluatorImpl(Authorizer authorizationChecker,
+      AuthorizationValidator authorizationValidator) {
+    this.authorizedPredicate = auth -> authorizationChecker.isAuthorized(auth.toString());
+    this.authorizationValidator = authorizationValidator;
```
dlmarion: Can the validator be null?

core/src/main/java/org/apache/accumulo/access/impl/AccessEvaluatorImpl.java
```diff
@@ -143,20 +157,29 @@ public boolean canAccess(AccessExpression expression) {
...
 
-  boolean evaluate(byte[] accessExpression) throws InvalidAccessExpressionException {
-    var bytesWrapper = ParserEvaluator.lookupWrappers.get();
+  boolean evaluate(String accessExpression) throws InvalidAccessExpressionException {
+    var charsWrapper = ParserEvaluator.lookupWrappers.get();
     Predicate<Tokenizer.AuthorizationToken> atp = authToken -> {
-      bytesWrapper.set(authToken.data, authToken.start, authToken.len);
-      return authorizedPredicate.test(bytesWrapper);
+      var authorization = ParserEvaluator.unescape(authToken, charsWrapper);
+      if (!authorizationValidator.test(authorization, authToken.quoting)) {
```
dlmarion: This will throw an NPE if the validator is null.

core/src/main/java/org/apache/accumulo/access/impl/AccessExpressionImpl.java
dlmarion: I wonder if this can be a Record and if there is any benefit to it.

core/src/main/java/org/apache/accumulo/access/impl/AccessEvaluatorImpl.java
keith-turner: It should not be, added test and checks in 6752d657d5a4e926784150575779e04b73edb8f5.  The checks for null are in the API entry point.

core/src/main/java/org/apache/accumulo/access/impl/AuthorizationsImpl.java
keith-turner: made record in 55a0c263f3bc6d97bfa97cc6ed92d19e1f658806

core/src/main/java/org/apache/accumulo/access/impl/CharsWrapper.java
```diff
@@ -18,46 +18,59 @@
...
-import static java.util.Objects.checkFromIndexSize;
-import static java.util.Objects.checkIndex;
-
 import java.util.Arrays;
+import java.util.Objects;
 
-public final class BytesWrapper implements Comparable<BytesWrapper> {
-
-  private byte[] data;
+public final class CharsWrapper implements CharSequence {
```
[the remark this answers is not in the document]
keith-turner: This class is mutable, so can not be a record.

core/src/main/java/org/apache/accumulo/access/impl/AccessExpressionImpl.java
keith-turner: Not sure, it currently does not have equals(), hashCode(), or toString().  Also, it has a mutable field that is in atomic ref.  It could be a record, but the atomic ref feels like it goes against the spirit of a record.

